PR #15, opened 2020-11-11
Remove unnecessary cargo copying

chris13524:
It doesn't seem like this line did anything useful and my project built fine without it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,6 @@ WORKDIR app
 COPY . .
 # Copy over the cached dependencies
 COPY --from=cacher /app/target target
-COPY --from=cacher /usr/local/cargo /usr/local/cargo
 RUN cargo build --release --bin app
 
 FROM rust as runtime


LukeMathWalker:
Without that line `cargo` will download again the dependencies, on top of building your crate.

chris13524:
Thanks @LukeMathWalker!

Could this requirement be documented somehow? I would imagine other's coming along could make the same assumption when building their own Dockerfiles. Is the original cargo being modified somehow by cargo chef?

LukeMathWalker:
No, no dark magic at play here 😁 
I think we can a comment on top of that line to explain that it's there to avoid re-downloading the dependencies to make it clearer.

chris13524:
Would you mind elaborating on how that line avoids re-downloading the dependencies? I am not understanding how those are related.

LukeMathWalker:
`cargo` will look into $CARGO_HOME for the downloaded copy of every registry dependency specified in your manifest ( https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads ). When using Debian as base image, that's `/usr/local/cargo`. If they are not there, it will download them again.

chris13524:
Oh my bad, I was under the assumption that `/usr/local/cargo` was the `cargo` binary, not a folder! I get it now. Sorry for the unnecessary noise :skull:

PR #22, opened 2020-11-21
Add cargo build targets to pick up dev-dependencies

BenD0G:
Hi there, big fan of this project 😄 thought I'd share a change that has helped me out - building multiple targets, which picks up `dev-dependencies`. Happy to discuss further!

## Changes
1. Add some flags to `cargo chef cook`, which are passed directly to `cargo build`:
    1. `--benches`
    2. `--tests`
    3. `--examples`
    4. `--all-targets`
2. Add skeleton files for `tests` and `examples` in the same fashion as already happens for `benches`.

## Motivating Example
In the build-test-release cycle, building and testing rust code often shares some (but not all) cacheable dependencies. With this change you can pre-compile both your production and your test dependencies, speeding up both, like so:
```dockerfile
FROM rust AS menu-planner
WORKDIR /app
COPY . /app/
RUN ["cargo", "chef", "prepare"]

FROM rust AS kitchen
WORKDIR /app
COPY --from=menu-planner /app/recipe.json /app/recipe.json
RUN ["cargo", "chef", "cook", "--release", "--all-targets"]

FROM rust AS builder
WORKDIR /app
COPY . /app/
COPY --from=kitchen /app/target /app/target
RUN ["cargo", "build", "--release", "--all-targets"]

FROM rust AS testtime
WORKDIR /app
COPY . /app/
COPY --from=builder /app/target /app/target
CMD ["cargo", "test", "--release"]

FROM rust AS runtime
COPY --from=builder /app/target/release/cachetest cachetest
CMD ["./cachetest"]
```
Then you can run `docker build . -t tester --target=testtime` to make full use of the cached dependencies.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -71,16 +71,17 @@ pub struct Cook {
     /// Space or comma separated list of features to activate.
     #[clap(long, use_delimiter = true, value_delimiter = ",")]
     features: Option<Vec<String>>,
-    /// Build benchmarks
+    /// Build all benches
     #[clap(long)]
     benches: bool,
-    /// Build tests
+    /// Build all tests
     #[clap(long)]
     tests: bool,
-    /// Build examples
+    /// Build all examples
     #[clap(long)]
     examples: bool,
-    /// Build all targets
+    /// Build all targets.
+    /// This is equivalent to specifying `--tests --benches --examples`.
     #[clap(long)]
     all_targets: bool,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 use anyhow::Context;
-use chef::{DefaultFeatures, OptimisationProfile, Recipe};
+use chef::{DefaultFeatures, OptimisationProfile, Recipe, TargetArgs};
 use clap::Clap;
 use fs_err as fs;
 use std::collections::HashSet;
@@ -71,6 +71,18 @@ pub struct Cook {
     /// Space or comma separated list of features to activate.
     #[clap(long, use_delimiter = true, value_delimiter = ",")]
     features: Option<Vec<String>>,
+    /// Build benchmarks
+    #[clap(long)]
+    benches: bool,
+    /// Build tests
+    #[clap(long)]
+    tests: bool,
+    /// Build examples
+    #[clap(long)]
+    examples: bool,
+    /// Build all targets
+    #[clap(long)]
+    all_targets: bool,
 }
 
 fn _main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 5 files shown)


LukeMathWalker:
Hey @BenD0G!
Thanks for the PR - I'll have look over the weekend most likely :grin:

src/main.rs
```diff
@@ -71,6 +71,18 @@ pub struct Cook {
     /// Space or comma separated list of features to activate.
     #[clap(long, use_delimiter = true, value_delimiter = ",")]
     features: Option<Vec<String>>,
+    /// Build benchmarks
```
LukeMathWalker, line 74:
```suggestion
    /// Build all benches
```

src/main.rs
```diff
@@ -71,6 +71,18 @@ pub struct Cook {
...
+    #[clap(long)]
+    benches: bool,
+    /// Build tests
```
LukeMathWalker, line 77:
```suggestion
    /// Build all tests
```

src/main.rs
```diff
@@ -71,6 +71,18 @@ pub struct Cook {
...
+    #[clap(long)]
+    tests: bool,
+    /// Build examples
```
LukeMathWalker, line 80:
```suggestion
    /// Build all examples
```

src/main.rs
```diff
@@ -71,6 +71,18 @@ pub struct Cook {
...
+    #[clap(long)]
+    examples: bool,
+    /// Build all targets
```
LukeMathWalker, line 83:
```suggestion
    /// Build all targets.
    /// This is equivalent to specifying `--tests --benches --examples`.
```

src/skeleton.rs
```diff
@@ -144,6 +144,36 @@ impl Skeleton {
...
+                let test_name = test.name.as_ref().context("Missing test name.")?;
+                let test_relative_path = test
+                    .path
+                    .clone()
+                    .unwrap_or_else(|| format!("tests/{}.rs", test_name));
+                let test_path = parent_directory.join(test_relative_path);
+                if let Some(parent_directory) = test_path.parent() {
+                    fs::create_dir_all(parent_directory)?;
+                }
+                fs::write(test_path, "fn main() {}")?;
```
LukeMathWalker, line 159:
```suggestion
                fs::write(test_path, "")?;
```

LukeMathWalker:
Thanks for the PR!
I modified a few minors things:
- help text in line with `cargo`'s, with the exception of `all-targets`;
- tests do not need a `main` function, an empty file will do just fine.

PR #28, opened 2020-12-02
Replace parsed build args with positional args

mwkroening:
This removes manual parsing of all `cargo-build` arguments. Instead, build arguments are now accepted through last positional arguments. Those arguments which are important for decisions made by `cargo-chef` are parsed from there.

This allows specifying all possible build arguments while still reacting to some of them. Originally, I was trying to pass `--workspace`, which is not possible currently, but found that it might be easier to just allow specifying any build argument instead of parsing each possible build argument explicitly.

What do you think?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 mod recipe;
 mod skeleton;
 
-pub use recipe::{DefaultFeatures, OptimisationProfile, Recipe, TargetArgs};
+pub use recipe::{OptimisationProfile, Recipe};
 pub use skeleton::*;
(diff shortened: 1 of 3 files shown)


LukeMathWalker:
Parsing the build arguments has been an intentional design choice in `cargo-chef` - certain build arguments require `cargo-chef` to perform additional actions and I prefer to have a dev opening an issue here saying "flag xyz is not supported" than having the tool behave weirdly because they hit a edge case I didn't account for.
Therefore I'll close this PR but I am happy to merge one adding support for `--workspace` :+1:

PR #29, opened 2020-12-03
Update Readme with more efficient planning stage example

Lesstat:
Hi,
I just started my first experiments with cargo-chef (as described in the readme) and noticed, that all my dependencies were rebuild when I changed my code.
In the Readme  the prepare stage is displayed like this:

```dockerfile
FROM rust as planner
WORKDIR app

RUN cargo install cargo-chef 
COPY . .
RUN cargo chef prepare  --recipe-path recipe.json
``` 
This means that the prepare stage depends on all files in the current directory. When we change this to:
```dockerfile
FROM rust as planner
WORKDIR app

RUN cargo install cargo-chef 
COPY Cargo.toml Cargo.lock /app/ # <-- this is different
RUN cargo chef prepare  --recipe-path recipe.json
``` 
This is no longer the case and it worked for my project. Unfortunately, I am not clear enough on the internals of cargo-chef and if this works for every project. If so I would propose to merge this pull request, otherwise feel free to close it :).

BTW: Thank you very much for this awesome tool!

Best regards,
Lesstat

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,8 @@ FROM rust as planner
 WORKDIR app
 # We only pay the installation cost once, 
 # it will be cached from the second build onwards
-RUN cargo install cargo-chef 
-COPY . .
+RUN cargo install cargo-chef
+COPY Cargo.toml Cargo.lock /app/
 RUN cargo chef prepare  --recipe-path recipe.json
 
 FROM rust as cacher


LukeMathWalker:
The `prepare` stage indeed depends on all file in your project, but the `prepare` does not perform any dependency building - it just outputs a `recipe.json` file. Apart from installing `cargo-chef` itself, which is going to be cached after the first build, it should take a few seconds at most. 
Being able to copy your working directory wholesale, without having to pick and choose, but still getting caching is a huge chunk of the value proposition here, so I don't think we should merge this.

Can you share the entire Dockerfile you were using where you saw this problem?

Lesstat:
I totally understand your reasoning. I think my problem came from a bad interaction with a `build.rs`-script which I can not exactly pinpoint.   

The dockerfile looked like this:
```dockerfile
FROM rust:1.48 as planner
WORKDIR app
# We only pay the installation cost once,
# it will be cached from the second build onwards
RUN cargo install cargo-chef
COPY . .
RUN cargo chef prepare  --recipe-path recipe.json

FROM rust:1.48 as cacher
WORKDIR app
RUN cargo install cargo-chef
COPY --from=planner /app/recipe.json recipe.json
RUN cargo chef cook --release --recipe-path recipe.json

FROM rust:1.48 as builder
WORKDIR app
COPY . .
# Copy over the cached dependencies
COPY --from=cacher /app/target target
COPY --from=cacher /usr/local/cargo /usr/local/cargo
RUN cargo build --release 

FROM scratch as runtime
WORKDIR app
COPY --from=builder /app/target/release/osmgraphing .
ENTRYPOINT ["/app/osmgraphing"]
```

LukeMathWalker:
Can you upload on GitHub a skeleton project that reproduces the issue you were seeing?

Lesstat:
I tried recreating my issues with a minimal project but unfortunately did not succeed.  If I manage do it sometime I will contact you again.

PR #30, opened 2020-12-07
Use environment variable instead of hardcoded value of $CARGO_HOME in readme

agersant:
Many thanks for working on cargo chef!

I was trying to adapt the example for my own use case and ran into this issue. I don't think the cargo home directory is guaranteed to be `/usr/local/cargo` so using [$CARGO_HOME](https://doc.rust-lang.org/cargo/guide/cargo-home.html) should make the example more portable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 2. [Benefits vs Limitations](#benefits-vs-limitations)
 3. [License](#license)
 
-## How To Install 
+## How To Install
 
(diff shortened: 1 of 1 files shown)


agersant:
I just realized my editor also stripped out all the trailing white-space, please let me know if you want it back!

LukeMathWalker (approved):
Definitely better this way, thank you!

Issue #31, opened 2020-12-09
cargo chef cook altering Cargo.toml and breaking projects in workspace


pythoneer:
## EDIT:

I realized that issue #27 is absolutely relevant here. I think its all just fine the way things are working and i guess i understand the reasons behind it – delete all the "code" and just fill the target folder with the compiled dependencies. The "real" code comes back later in the build stage. Silly me. I just tried it with a little project and it worked (just using the docker commands) and i wanted it to extend to a more complex one and in the middle of development "take a look" at what cargo-chef would do to see if there are some problems already i need to fix before going the full docker route. 

And now that i think i know how this hole thing is working – i feel a little bit stupid. But i agree with the comments in the other issue that a warning would be helpful to make the people using it aware that a) you're not supposed to use it "locally" and b) it will delete and alter your files on purpose and maybe with an explanation why. I'll try to get my test project working without "test cooking" locally and use it like it is supposed to :P 

---- original message

I think this is related to issue #4 but i have the additional problem that `cargo chef cook` is also breaking my project by altering the respective Cargo.toml files from each project under a workspace. I have prepared a test project for you to replicate https://github.com/pythoneer/docker_build_test

`cargo make build` runs fine
`cargo chef prepare  --recipe-path recipe.json` looks like its doing OK (can't tell if the *.json is correct)
`cargo chef cook --recipe-path recipe.json` breaks the project and
`cargo make build` does not build the project anymore because the Cargo.toml files are just wrong now

I don't know if this is not supposed to work in the first place because i haven't found a hint in the README (maybe the section about `Limitations and caveats` but to be honest i don't really understand that 100%)

I hope this helps. If you need any more information i am glad to help. 

(love your book :+1: )

pythoneer:
Sry, i now realized that `cargo chef cook` is also deleting the content project files like `shared/src/lib.rs` and `server/src/main.rs` .. therefore my test project is currently broken ... i'll fix it later because i am unfortunately on my way now.

LukeMathWalker:
This is by design, as discussed in #4 - `cargo-chef` is designed for usage in containers to leverage Docker layer caching, where changes to the source are considered ephemeral given that the container itself is ephemeral.

Some of the improvements discussed in #4 should hopefully avoid this surprise in the future.

pythoneer:
Thanks for you answer i think that concludes my journey here :) ... i'll close this issue as it is redundant and almost a duplicate from issue  #4 / #27

Issue #27, opened 2020-12-02
Chef seems to "empty" crates entrypoints


MatteoJoliveau:
Hi!
I tried running `cargo chef cook` on [Enseada](https://github.com/enseadaio/enseada), which is composed of several crates linked up into a `server` crate that contains the `main` module.

After running `cook`, all the `lib.rs` become completely empty, and the `main.rs` only contains the following:

```rust
fn main() {}
```

I really don't think this is the desired behaviour 😄 

I was able to reproduce it with a fresh repo: https://github.com/MatteoJoliveau/cargo-chef-repro
The same apply to a repo without dependencies (e.g. a bare repo generated with `cargo new <name>`)

It also modifies the `Cargo.toml` which I really would like to keep the way I wrote it 😅

MatteoJoliveau:
Running cargo-chef 0.1.9 btw

MatteoJoliveau:
By looking a the code I found the function `build_minimum_project` that is invoked by `chef cook` and that requires to be run on a blank canvas. 
Does this mean we should not be using chef locally, or on an existent codebase?
I guess it's fine for building container images but it would have been useful to speedup local development too

LukeMathWalker:
Hey!
`cargo-chef` is indeed designed to be used in containers. I don't see the usecase for local development - what would it give you compared to something like incremental cargo compilation?

MatteoJoliveau:
I guess you're right, I often run `cargo clean` as a habit from other languages where previous compilation results may interfere with the current one, but I guess it's not really necessary given how well cargo works.

Maybe we should document this behaviour nonetheless? Either point out that chef is not really meant to be run outside container builds and/or that it should be run on a clean directory structure.

simon-bourne:
I ran into this as well, when initially trying out `cargo-chef`. While it's not a problem once you figure out how it works, it's a bit surprising while experimenting.

Would it be better to create empty files if they don't exist, or check they are empty if they do exist? This way you can run `cook` on a fresh directory, or re-run it on the same directory (for example to cache debug and release variants), but you'd never overwrite any code.

LukeMathWalker:
I think that's a very good idea @simon-bourne - `cargo chef cook` should exit with a meaningful user warning if it needs to overwrite existing files.

MatteoJoliveau:
[...]

PR #32, opened 2020-12-10
Warning about file overwrite when stdout is a TTY

MatteoJoliveau:
This PR addresses #27 and #31 by introducing a warning and interactive prompt when stdout is detected as being a TTY.

This should avoid surprise and actual source files being overwritten by the skeleton when running `cook` outside of container builds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,6 +81,7 @@ dependencies = [
  "anyhow",
  "assert_cmd",
  "assert_fs",
+ "atty",
  "cargo-manifest",
  "clap",
  "env_logger",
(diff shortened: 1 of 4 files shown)


MatteoJoliveau:
@LukeMathWalker I'm not sure about how to test this change. Is there a good way to test stdin/stdout features running in `fn main()`?

LukeMathWalker:
You can definitely test `stdout` using [`assert_cmd`](https://github.com/assert-rs/assert_cmd) - not sure about "dynamic" stdin prompts.

LukeMathWalker (approved):

PR #33, opened 2020-12-15
Fixed ordering of autobins.

markhildreth:
When there are auto binaries (binaries not listed in the Cargo.toml, but
found in the "src/bin" directory), the only way to pick those up is
using the call to `cargo_manifest::Manifest::complete_from_path`.
However, there is no guarantee made here to the order of the binaries
after this call. As a result, it's often the case that the ordering of
the binaries is not consistent.

The skeleton is derived from this manifest, and the resulting recipe
file will differ because the binaries are in a different order. This can
cause Docker images to rebuild because they see a different file. In
reality, the data is the same, but in a sightly different order.

This commit sorts the order of the binaries in the manifest structure
before the file is written to disk, working around this problem.

### DIFF
--- a/src/skeleton.rs
+++ b/src/skeleton.rs
@@ -40,7 +40,30 @@ impl Skeleton {
                     // As suggested in issue #142 on toml-rs github repository
                     // First convert the Config instance to a toml Value,
                     // then serialize it to toml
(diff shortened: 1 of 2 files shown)


LukeMathWalker (approved):
Thanks a lot for this!

PR #44, opened 2021-01-18
cook: clean up also dummy build.rs artifacts

### LINKED ISSUE #42: Presence of build.rs breaks build with cargo chef: generated files not present
Hi, I realise the README says
> cargo-chef has not yet been tested extensively with projects leveraging build files.

...this issue is an attempt to change that. :)

I have created a minimal example crate leveraging build.rs file: https://github.com/strohel/cargo-chef-build-rs

The build using cargo-chef fails with the following error:
```
Step 16/20 : RUN cargo build --release --bin cargo-chef-build-rs
 ---> Running in 53c5f9739b5f
   Compiling cargo-chef-build-rs v0.1.0 (/app)
error: couldn't read /app/target/release/build/cargo-chef-build-rs-087b5874e2eb29d7/out/built.rs: No such file or directory (os error 2)
 --> src/main.rs:1:1
  |
1 | include!(concat!(env!("OUT_DIR"), "/built.rs"));
```
i.e. the file that is generated by `build.rs` is not present. It is kinda strange, because at this point, cargo chef is no longer involved (beyond files generated by it being present in /target). Maybe that somehow fools `cargo` to think that `build.rs` was already run, while it actually wasn't?

I'll be happy to help to resolve this.

strohel:
Should fix #42.

Fixes a build of an example project https://github.com/strohel/cargo-chef-build-rs for me.

We only needed to remove the case 1. b). That makes sense, because fingerprints only check what is in `build/`, and run of the build script has a dependency on its build.

Note that there is special case, which currently builds, but is not fully cached: when there is a `build.rs` file on filesystem, but no mention of `build = "build.rs"` in Cargo.toml. This one is separate though. It's proper fix would be to extend `cargo-manifest` to also autodetect `build` property in its `Manifest::complete_from_abstract_filesystem()` method (in addition to bins, libs, ...).

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -43,7 +43,7 @@ impl Recipe {
             target_args,
         );
         self.skeleton
-            .remove_compiled_dummy_libraries(current_directory, profile, target, target_dir)
+            .remove_compiled_dummies(current_directory, profile, target, target_dir)
             .context("Failed to clean up dummy compilation artifacts.")?;
         Ok(())
     }
--- a/src/skeleton.rs
+++ b/src/skeleton.rs
@@ -219,7 +219,7 @@ impl Skeleton {
     /// workspace.
     /// Given the usage of dummy `lib.rs` files, keeping them around leads to funny compilation
     /// errors if they are a dependency of another project within the workspace.
-    pub fn remove_compiled_dummy_libraries<P: AsRef<Path>>(
+    pub fn remove_compiled_dummies<P: AsRef<Path>>(
         &self,
         base_path: P,
         profile: OptimisationProfile,
--- a/src/skeleton.rs
+++ b/src/skeleton.rs
@@ -215,10 +215,10 @@ impl Skeleton {
         Ok(())
     }
 
-    /// Scan the target directory and remove all compilation artifacts for libraries from the current
-    /// workspace.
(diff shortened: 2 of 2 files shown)


LukeMathWalker:
I merged https://github.com/LukeMathWalker/cargo-manifest/pull/8, released `cargo-manifest` and updated `main` to use the latest version ✔️

strohel:
> I merged [LukeMathWalker/cargo-manifest#8](https://github.com/LukeMathWalker/cargo-manifest/pull/8), released `cargo-manifest` and updated `main` to use the latest version heavy_check_mark

Thanks a bunch! With that and this merged, `build.rs` support should be good -- if you also want to release cargo-chef, it would be awesome.

LukeMathWalker:
Updated the README and released - thanks a lot @strohel!

Palmik:
I think this might be breaking cases where you have a lib with `tonic::include_proto` as the artifacts generated by `build.rs` are not present in the build stage.

strohel:
@Palmik wrote:
> I think this might be breaking cases where you have a lib with tonic::include_proto as the artifacts generated by build.rs are not present in the build stage.

It would be cool if you tried whether that actually breaks something, and if so, describe the breakage in a new issue.

Note that this PR only affects `build.rs` in the final (top-level) crate, and building those didn't work at all with `cargo chef` before this PR.

PR #46, opened 2021-01-20
Add docker GitHub Actions workflow to build docker images 

### LINKED ISSUE #41: Provide Docker image with cargo-chef
Hi, it would be cool if you could provide a Docker image on Docker Hub with `cargo-chef` installed.

Its Dockerfile could be as simple as
```Dockerfile
FROM rust
RUN cargo install cargo-chef 
```
The README is right that the `RUN cargo install cargo-chef` step gets cached, but that only holds if you run it on the same machine every time. A lot of CIs run each job on a different (virtual) machine, with empty local Docker layer cache.

(it is possible to cache using remote images with --cache-from, --target, and manual pushes/pulls, but it's less convenient)

Thanks for a very helpful tool.

strohel:
Currently uses just the "latest" tag of the upstream rust base image, but the
approach is extensible to many such tags.

Requires following GitHub repository secrets to be defined:

DOCKERHUB_USERNAME - username on https://hub.docker.com/
DOCKERHUB_TOKEN - token from Docker Hub for DOCKERHUB_USERNAME used to push images.

The job is scheduled to run every night, exiting early if current images are
up-to-date.

Seems to work fine:
- built image: https://hub.docker.com/r/strohel/cargo-chef
- GitHub job that built the image: https://github.com/strohel/cargo-chef/runs/1735587797
- GitHub job that was no-op because versions were up-to-date: https://github.com/strohel/cargo-chef/runs/1735685360

Fixes #41.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,44 @@
+name: Build Docker images
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/docker.yml'
+      - 'docker/**'
+  schedule:
+    - cron: '42 7 * * *' # run at 7:42 UTC (morning) every day
+jobs:
+  build:
+    name: Build and push
+    runs-on: ubuntu-latest
+    env:
+      RUST_IMAGE_TAG: latest
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          docker login --username $DOCKERHUB_USERNAME --password $DOCKERHUB_TOKEN
+
+          RUST_IMAGE=rust:$RUST_IMAGE_TAG
+          RUST_VERSION=$(docker run --rm $RUST_IMAGE rustc --version)
+          echo "Rust version in $RUST_IMAGE is $RUST_VERSION."
+
+          CHEF_IMAGE=$DOCKERHUB_USERNAME/cargo-chef:$RUST_IMAGE_TAG
+          CHEF_RUST_VERSION=$(docker run --rm $CHEF_IMAGE rustc --version || echo "none")
+          echo "Rust version in $CHEF_IMAGE is $CHEF_RUST_VERSION."
+          CHEF_VERSION=$(docker run --rm $CHEF_IMAGE cargo install --list | grep "^cargo-chef v" | cut -d v -f 2 | cut -d : -f 1 || echo "none")
+          echo "cargo chef version in $CHEF_IMAGE is $CHEF_VERSION."
+
+          # Somewhat tricky way to determine latest version without installing anything
+          LATEST_CHEF=$(cargo search --limit 1 cargo-chef | head -n 1 | cut -d '"' -f 2)
+          echo "Latest cargo chef version on crates.io is $LATEST_CHEF."
+
+          if [ "$RUST_VERSION" == "$CHEF_RUST_VERSION" -a "$LATEST_CHEF" == "$CHEF_VERSION" ]; then
+            echo "All versions up-to-date, existing early."
+            exit 0
+          fi
+
+          docker build docker/ -t $CHEF_IMAGE --build-arg=BASE_IMAGE=$RUST_IMAGE
+          docker push $CHEF_IMAGE
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,4 @@
+ARG BASE_IMAGE=rust
+FROM $BASE_IMAGE
+
+RUN cargo install cargo-chef && rm -rf $CARGO_HOME/registry/
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,5 @@
+This directory contains Dockefile to build cargo-chef image that can be used as
+a base layer for `planner` and `cacher` build stages.
+
+Note that this builds `carcho chef` from latest release on crates.io - not from
+the current repository.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
+name: Build Docker images
+on:
+  push:
+    branches:
+      - main
```
strohel, line 5: Uncomment these to test before merging to the `main` branch.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
...
+
+          RUST_IMAGE=rust:$RUST_IMAGE_TAG
+          docker run --rm $RUST_IMAGE rustc --version
+          RUST_VERSION=$(docker run --rm $RUST_IMAGE rustc --version)
+          echo "Rust version in $RUST_IMAGE is $RUST_VERSION."
+
+          CHEF_IMAGE=$DOCKERHUB_USERNAME/cargo-chef:$RUST_IMAGE_TAG
+          CHEF_RUST_VERSION=$(docker run --rm $CHEF_IMAGE rustc --version || echo "none")
+          echo "Rust version in $CHEF_IMAGE is $CHEF_RUST_VERSION."
+          CHEF_VERSION=$(docker run --rm $CHEF_IMAGE cargo install --list | grep "^cargo-chef v" | cut -d v -f 2 | cut -d : -f 1 || echo "none")
```
strohel, line 32: This is a bit cheesy, but `cargo chef --version` currently returns a dummy value.

Even if it did, the script would depend on the fact that it returns _exactly same_ value as `cargo search cargo-chef` - i.e. any fancy stuff like `1.49.0 (e1884a8e3 2020-12-29)` would break it.

strohel:
When this is merged and confirmed to work, the README should be updated to point people to using the image, rather than building cargo chef themselves.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
...
+      RUST_IMAGE_TAG: latest
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          docker login --username $DOCKERHUB_USERNAME --password $DOCKERHUB_TOKEN
+
+          RUST_IMAGE=rust:$RUST_IMAGE_TAG
+          docker run --rm $RUST_IMAGE rustc --version
```
TriplEight, line 25: GitHub Action's `ubuntu-latest` has Rust built-in, so there's no need to use a container for it.

.github/workflows/docker.yml
TriplEight: If this job would run on releases, i.e.
```
on:
  release:
    types:
      - published
```
then it would be possible to use `${{ github.event.release.tag_name }}` as a version tag for an image. (That also would be a fix for the dummy `--version` output).

.github/workflows/docker.yml
strohel: This line is actually a leftover from my debugging - I've removed it. I've noticed it thanks to your comment.

The next line `RUST_VERSION=$(docker run ...` needs to use rustc from `$RUST_IMAGE`, because its version needs to match rustc from $CHEF_IMAGE (which is itself based on $RUST_IMAGE).

.github/workflows/docker.yml
```diff
@@ -0,0 +1,44 @@
...
+  build:
+    name: Build and push
+    runs-on: ubuntu-latest
+    env:
+      RUST_IMAGE_TAG: latest
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
```
TriplEight, line 21: There's an official "docker build and push action" https://github.com/marketplace/actions/build-and-push-docker-images
And also buildah actions, for the connoisseurs: https://github.com/marketplace/actions/buildah-build and https://github.com/redhat-actions/push-to-registry

.github/workflows/docker.yml
LukeMathWalker: The job will have to run a schedule to release a new version every time Rust releases, even if cargo-chef did not cut a new release.
But @TriplEight makes a very good point - we can rely on the latest non-alpha tag in the repository to detect the latest version of `cargo-chef`, no need to touch Docker.

LukeMathWalker: For the Rust version - we could use a [label](https://docs.docker.com/engine/reference/builder/#label)

TriplEight: Sorry, I didn't quite get the idea behind running this job on schedule for every new Rust. Can you elaborate, please? 
You do not even test the code against `nightly`. If it would be the case, and if you need to release a container for every push, just place this job right after one which would run `cargo +nightly test ...`.

strohel: > The job will have to run a schedule to release a new version every time Rust releases, even if cargo-chef did not cut a new release.

Right, I was about to say exactly this.

> But @TriplEight makes a very good point - we can rely on the latest non-alpha tag in the repository to detect the latest version of cargo-chef, no need to touch Docker.

Well, this particular command here is to query the version of cargo chef _in the latest published docker image_ (not latest version published to crates.io).

.github/workflows/docker.yml
strohel: Not sure whether you suggest something particular, but I'd be curious to see an alternative PR implementing the required logic using one of these pre-defined actions - "patches welcome" I guess. :) The tricky part is not to build when there's no need to (unless we want to waste GitHub's resources).

TriplEight: Maybe I just didn't understand why creating this on schedule to follow the latest Rust exactly in the containerization job.

.github/workflows/docker.yml
LukeMathWalker: > Sorry, I didn't quite get the idea behind running this job on schedule for every new Rust. Can you elaborate, please?

We are referring to stable Rust releases.
E.g. if the latest cargo chef is `0.1.12` and we build the Docker image today, the released Docker image will be using Rust `1.49.0`. When Rust `1.50.0` is released, we want to release a new image with cargo-chef `0.1.12` and Rust `1.50.0`.

strohel: > Sorry, I didn't quite get the idea behind running this job on schedule for every new Rust. Can you elaborate, please?

It is somewhat briefly described in #41 and its comments. The idea is to have have up-to-date images with latest released cargo-chef, eventually for multiple tags of the official `rust` image, but starting with just `latest` in this PR (but a matter of setting one variable). Notice that release cadence of `rust` (including minor versions) may be actually higher than of `cargo chef`, hence a need for a schedule.

> For the Rust version - we could use a label

Yes it would be technically possible to label $CHEF_IMAGE (though we have no power to label $RUST_IMAGE), and then read it (which would still AFAICS require pulling the image). But in order to label it, we would still need to determine the version (or rust in $RUST_IMAGE) somehow? Maybe that's just a lack of my imagination - feel free to suggest concrete changes to code.

LukeMathWalker:
Merged, updated the README and triggered the first build successfully: https://hub.docker.com/repository/docker/lukemathwalker/cargo-chef# 😁
Thanks a lot @strohel!

I have only modified the triggers: it now builds on a schedule + on every tag (so that we trigger a build every time cargo-chef is released).

PR #52, opened 2021-03-17
Remove old incorrect commit

deantvv:
Issue #142 on toml-rs is already solved and is no longer the reason to parse to toml `Value`. 
The reason why this is needed is to sort the manifest.


LukeMathWalker (approved):
Thanks!

Issue #35, opened 2020-12-20
Changing project version, without changing dependency tree, modifies recipe.json and invalidates cache.


quebin31:
Is this actually the expected behavior?  From my point of view, changing the project version shouldn't affect the overall build of dependencies, but in this case, this information is included in the `recipe.json` file, and even if you don't change the dependency tree this triggers a full rebuild of dependencies since the cache has been invalidated due to the change.

LukeMathWalker:
It definitely should not happen :sweat_smile: 
It should not be too hard to fix, but it might require fiddling around a bit with the manifests.

LukeMathWalker:
Fixed by #53 thanks to @deantvv

PR #53, opened 2021-03-17
Ignore package.version to avoid unnecessary rebuild

deantvv:
In the past, package.version is track in `recipe.json` which will trigger full
project rebuild if package.version is bumped. To fix this issue, this patch
replace package.version to a constant value, so it won't trigger full project
rebuild.

Fix issue #35


LukeMathWalker (approved):

LukeMathWalker:
Thanks 😁

PR #57, opened 2021-04-08
Add support for --workspace

mkroening:
I would like to cook a whole workspace, but would prefer not to specify each member via `--package` manually.

This PR introduces the corresponding `--workspace` argument (superseding #28).

What do you think?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,6 +91,9 @@ pub struct Cook {
     /// Package to build (see `cargo help pkgid`)
     #[clap(long, short = 'p')]
     package: Option<String>,
+    /// Build all members in the workspace.
+    #[clap(long)]
+    workspace: bool,
 }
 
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
Isn't this the default behaviour if you are running `cargo chef` in a workspace?

mwkroening:
Aha, I dug deeper and gained some insight.

Our project has a workspace with a root package. `cargo-chef` does not correctly parse `default-members`, which selects which packages are affected by `cargo build` by default. Without `default-members`, only the root package is built. Coincidentally this issue should be fixed by https://github.com/LukeMathWalker/cargo-manifest/pull/9, which I originally opened because of a similar issue with `proc-macro`.

Generally speaking though, it might be useful to provide the option to cook all packages of the workspace, not just the default ones.

LukeMathWalker:
Sounds good, thanks for working on it 😁

PR #61, opened 2021-05-07
Included refresh instructions in README

panasenco:
Thank you so much for this tool! I use it to efficiently build the [Scryer Prolog Docker image](https://github.com/mthom/scryer-prolog/blob/master/Dockerfile). I was completely lost dealing with complicated dependencies before I found `cargo-chef `.

I added a sentence about refreshing the cacher to the README. Might save someone a head scratching moment and a search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ COPY --from=builder /app/target/release/app /usr/local/bin
 ENTRYPOINT ["/usr/local/bin/app"]
 ```
 
+When your cacher becomes stale (you start seeing the builder stage rebuild everything), refresh the cacher by passing the `--no-cache` flag to `docker build`.
+
 ## Benefits vs Limitations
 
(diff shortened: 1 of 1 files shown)


panasenco:
Actually the builder still tries to rebuild everything even after clearing the builder cache so I don't know what's going on...

panasenco:
It worked after switching from your cargo-chef image to the official rust image. Maybe the rust version in your image is out of date? That or Docker build cache is simply weird.

LukeMathWalker:
The latest cargo-chef version is on Rust 1.52.0 - see https://github.com/LukeMathWalker/cargo-chef/runs/2537999233?check_suite_focus=true
What might have happened is the following: your first layers were cached, so they kept using an older version of the image (probably on Rust 1.51.0) which then triggered a full rebuild in the builder stage that was running on Rust 1.52.0

PR #65, opened 2021-06-01
Add `--offline` option

prestontw:
to support building with vendored and patched dependencies. 

I typically run my build with
```
cargo build --offline -p <specific package name>
```
I tried translating my existing Dockerfile to use cargo chef,
but I was running into compilation errors where it wasn't
picking up the patched dependencies.

I commented out the portion of my Dockerfile
that ran `cook` and instead ran the above command.
This indeed works. So, I've added this option to the
necessary structs and where the `cargo` command is built.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -94,6 +94,9 @@ pub struct Cook {
     /// Build all members in the workspace.
     #[clap(long)]
     workspace: bool,
+    /// Build offline.
+    #[clap(long)]
+    offline: bool,
 }
 
 fn _main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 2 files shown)


prestontw:
@LukeMathWalker I tried to add you as a reviewer, but it doesn't seem like it's available? Please let me know if there's anything I need to do to contribute!

prestontw:
Another option besides adding more options to cargo chef is to just pass through all arguments besides the recipe path to cargo when we run the build step.

LukeMathWalker:
Thanks for the PR @prestontw!

As mentioned in the discussion you linked, I don't want to allow the user to pass raw flags/options to `cargo build`. If you end up needing other knobs, just open a PR/issue to discuss :)

PR #67, opened 2021-06-02
Ptw/pass remaining args through

prestontw:
This is an alternate approach to https://github.com/LukeMathWalker/cargo-chef/pull/65.
Instead of establishing option parity with `cargo`, let's just put all options that we don't
use for cleaning up the skeleton directory into a separate `cargo_options` field.

Note! If someone includes an option that we want, say `profile` or `recipe_path`, after an
unknown option, we would include these options for `cargo chef` only for cargo:
```shell
cargo chef --package test --recipe-path ./recipe.json
```
would lead to `./recipe.json` being run in `cargo build`, which doesn't make any sense.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 mod recipe;
 mod skeleton;
 
-pub use recipe::{CookArgs, DefaultFeatures, OptimisationProfile, Recipe, TargetArgs};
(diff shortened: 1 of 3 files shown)


prestontw:
I see that a similar approach was closed in https://github.com/LukeMathWalker/cargo-chef/pull/28. I'll close this out!

Issue #64, opened 2021-06-01
Feature request: `cargo-chef` support for patches


prestontw:
Howdy! I work in a medium-large workspace that includes [patches](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) to a couple of dependencies (can provide a minimal example if that would be helpful). When trying to use `cargo chef cook`, it doesn't seem like the patches are included or applied---I'm seeing compilation errors.

Would it be possible to add [rust-patches](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) support to `cargo-chef`?

prestontw:
I think the only thing that is necessary is adding `--offline` support to the cargo command that is run, which I've attempted to do in https://github.com/LukeMathWalker/cargo-chef/pull/65

prestontw:
I am still running into problems even with `--offline` support. I think it's because `cargo chef` is overwriting the rust-patches' `lib.rs` files. These should still be treated as dependencies---they should not be overwritten. I verified locally that `lib.rs` is getting replaced with an empty file. My first thought would be to allow a list of directories not to replace, and make the default include `rust-patches`. I'll try this locally and see if this solves the problem.

LukeMathWalker:
A minimal example would really help here to understand what the issue is and the best way forward to solve it (if possible).

prestontw:
@LukeMathWalker I've verified that https://github.com/prestontw/cargo-chef/commit/582504172902260d648e701431d252ebc925c579 fixes the problem for me. I'm planning on making this customizable and reproducing a minimal example. I wanted to show you what worked first, though, in case you had thoughts.

PR #70, opened 2021-06-07
Ptw/do not skeleton patches

prestontw:
I've verified that this addresses https://github.com/LukeMathWalker/cargo-chef/issues/64 locally.
I've also added a test!

My main repository has several patched dependencies. `cargo chef` would turn these dependencies'
`lib.rs` files into [empty strings](https://github.com/LukeMathWalker/cargo-chef/blob/5f10f909d2eddaad02c22d9df2b7d953cecb9c8d/src/skeleton.rs#L180-L189).
This would cause compilation issues since code I had patched in was suddenly missing.

Edit: I've now taken the below alternative approach. ~To avoid erasing the bodies of these dependencies, I've added an `ignore_regexes` parameter
for `Prepare`. I would specify `... --ignore-regexes "rust-patches/*"`. It's important that I use neither `"./rust-patches/*"` since that would be treated as a regex nor `rust-patches/*` since that would be expanded by the shell.~

An **alternative approach** would be to get this information automatically from sections of the `Cargo.toml`
that contain `patch`. @LukeMathWalker what do you think of manually specifying directories to ignore
vs inferring them from `Cargo.toml`?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,9 +48,10 @@ pub struct Prepare {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
-
     /// Space or comma separated regular expressions to ignore. This is useful if you have sources in the directory
-    /// that are "like dependencies"---for example, patches.
+    /// that are "like dependencies"---for example, patches. If you want to ignore a top-level directory named `rust-patches`,
+    /// you would ignore it with:
+    /// `cargo chef prepare ... --ignore-regexes "rust-patches/*"`
     ///
     /// It defaults an empty list.
     #[clap(long, use_delimiter = true, value_delimiter = ",")]
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,6 +48,13 @@ pub struct Prepare {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
+
+    /// Space or comma separated regular expressions to ignore. This is useful if you have sources in the directory
+    /// that are "like dependencies"---for example, patches.
+    ///
+    /// It defaults an empty list.
+    #[clap(long, use_delimiter = true, value_delimiter = ",")]
+    ignore_regexes: Vec<String>,
 }
 
 #[derive(Clap)]
@@ -190,8 +197,12 @@ fn _main() -> Result<(), anyhow::Error> {
                 })
                 .context("Failed to cook recipe.")?;
         }
-        Command::Prepare(Prepare { recipe_path }) => {
-            let recipe = Recipe::prepare(current_directory).context("Failed to compute recipe")?;
+        Command::Prepare(Prepare {
+            recipe_path,
+            ignore_regexes,
+        }) => {
+            let recipe = Recipe::prepare(current_directory, &ignore_regexes)
+                .context("Failed to compute recipe")?;
             let serialized =
                 serde_json::to_string(&recipe).context("Failed to serialize recipe.")?;
             fs::write(recipe_path, serialized).context("Failed to save recipe to 'recipe.json'")?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,14 +48,6 @@ pub struct Prepare {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
-    /// Space or comma separated regular expressions to ignore. This is useful if you have sources in the directory
-    /// that are "like dependencies"---for example, patches. If you want to ignore a top-level directory named `rust-patches`,
-    /// you would ignore it with:
-    /// `cargo chef prepare ... --ignore-regexes "rust-patches/*"`
-    ///
-    /// It defaults an empty list.
-    #[clap(long, use_delimiter = true, value_delimiter = ",")]
-    ignore_regexes: Vec<String>,
 }
 
 #[derive(Clap)]
@@ -198,12 +190,8 @@ fn _main() -> Result<(), anyhow::Error> {
                 })
                 .context("Failed to cook recipe.")?;
         }
-        Command::Prepare(Prepare {
-            recipe_path,
-            ignore_regexes,
-        }) => {
-            let recipe = Recipe::prepare(current_directory, &ignore_regexes)
-                .context("Failed to compute recipe")?;
+        Command::Prepare(Prepare { recipe_path }) => {
+            let recipe = Recipe::prepare(current_directory).context("Failed to compute recipe")?;
             let serialized =
                 serde_json::to_string(&recipe).context("Failed to serialize recipe.")?;
             fs::write(recipe_path, serialized).context("Failed to save recipe to 'recipe.json'")?;
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -31,8 +31,8 @@ pub struct CookArgs {
 }
 
 impl Recipe {
-    pub fn prepare(base_path: PathBuf, ignore_regexes: &[String]) -> Result<Self, anyhow::Error> {
-        let skeleton = Skeleton::derive(&base_path, ignore_regexes)?;
+    pub fn prepare(base_path: PathBuf) -> Result<Self, anyhow::Error> {
+        let skeleton = Skeleton::derive(&base_path)?;
         Ok(Recipe { skeleton })
     }
 
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -31,8 +31,8 @@ pub struct CookArgs {
 }
 
 impl Recipe {
-    pub fn prepare(base_path: PathBuf) -> Result<Self, anyhow::Error> {
-        let skeleton = Skeleton::derive(&base_path)?;
+    pub fn prepare(base_path: PathBuf, ignore_regexes: &[String]) -> Result<Self, anyhow::Error> {
+        let skeleton = Skeleton::derive(&base_path, ignore_regexes)?;
         Ok(Recipe { skeleton })
     }
 
--- a/src/skeleton.rs
+++ b/src/skeleton.rs
@@ -31,7 +31,7 @@ impl Skeleton {
         let cargo_regex = "/**/Cargo.toml";
         let builder = if let Some(patches) = get_top_level_patches(&base_path) {
             let regex_patterns = std::iter::once(cargo_regex.to_string())
-                .chain(patches.iter().map(|s| String::from("!") + s))
+                .chain(patches.into_iter().map(|s| String::from("!") + &s))
                 .collect::<Vec<_>>();
             GlobWalkerBuilder::from_patterns(&base_path, &regex_patterns)
         } else {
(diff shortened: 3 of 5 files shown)


prestontw:
> An **alternative approach** would be to get this information automatically from sections of the `Cargo.toml`
> that contain `patch`. @LukeMathWalker what do you think of manually specifying directories to ignore
> vs inferring them from `Cargo.toml`?

I implemented this approach in d570ac4. I personally like it better since the patch information is already encoded in the `Cargo.toml` file and it avoids the two `"rust-patches/*"` issues above. A slight issue is that it only handles top-level `[patch.*]` entries, but maybe we can wait to handle that until someone actually needs it.

Let me know what you think! I'm happy to either revert it or try to add support for non-top-level `patch` entries.

prestontw:
@LukeMathWalker what do you think of the test I added? Do you think it's enough of a minimal example?

LukeMathWalker:
I have seen the PR but I haven't had time to review it yet @prestontw - I'll provide feedback here when I do :)

prestontw:
Here are what I see as the pro's and con's of each of the different approaches:

| Approach | Pro's | Con's |
| --- | --- | --- |
| Manually specifying via flags | The user has direct control over the paths that we do not "skelefy". This might be useful for other use cases outside of patches. | Duplicate information between the `Cargo.toml` files and `cargo chef` invocations. This ties our Dockerfiles to changes in `Cargo.toml`. |
| Exact path from Cargo.toml | Automatically updates which directories to leave out as people change the patches. | Forces users to list all directories in a dependency as a patch, even if they don't need it for the project themselves. For example, someone might need only a couple of [Rocket's crates](https://github.com/LukeMathWalker/cargo-chef/pull/70/files#diff-ae654e82cffe6f74b2b095b619a06eb9c86d63db01389c02c18f79a27d31fa02R372-R375), but they would have to add all of Rocket's sub directories as patches, *even though they don't need them*. |
| Parent of path in Cargo.toml | Avoids manual edits from the first row and specifying all directories from the second row. | Is a heuristic and might not work for everyone's project layout. |

Maybe a mix of the third and first row would be a good compromise? We try to infer the paths, but we give someone the option to override that list if they want.

LukeMathWalker:
The first row is not an option I am afraid - I want to maintain a minimal CLI interface for `cargo-chef` and I don't think this usecase is big enough to deserve bleeding into the user interface.

Can you go into more details on

> Forces users to list all directories in a dependency as a patch, even if they don't need it for the project themselves. For example, someone might need only a couple of Rocket's crates, but they would have to add all of Rocket's sub directories as patches, even though they don't need them.

I am not sure I understand what you mean here. On top of the test, could you provide an actual minimal reproducible example as a GitHub repo? It would make it easier to understand the current failure mode as well (you mentioned compilation errors, but you didn't provide details about them).

prestontw:
> (you mentioned compilation errors, but you didn't provide details about them).

Hi @LukeMathWalker, sorry for the delay. It's been hard to find time to try to minimize the problem, so in the meantime, here are some compilation issues I'm running into:
```
#39 36.70 error[E0433]: failed to resolve: could not find `signature` in `ring`
#39 36.70    --> /usr/local/cargo/registry/src/-f8edbd7a07552c5f/sct-0.5.0/src/lib.rs:165:40
#39 36.70     |
#39 36.70 165 |             RSA_PKCS1_SHA384 => &ring::signature::RSA_PKCS1_2048_8192_SHA384,
#39 36.70     |                                        ^^^^^^^^^ could not find `signature` in `ring`
#39 36.70 
#39 36.72 error[E0433]: failed to resolve: could not find `signature` in `ring`
#39 36.72    --> /usr/local/cargo/registry/src/-f8edbd7a07552c5f/sct-0.5.0/src/lib.rs:183:15
#39 36.72     |
#39 36.72 183 |         ring::signature::verify(alg, key, data, sig)
#39 36.72     |               ^^^^^^^^^ could not find `signature` in `ring`
#39 36.72 
#39 36.79 error: aborting due to 76 previous errors
```
We patched `ring` at version `0.14.6`. `cargo-chef` skelefies `ring`'s `lib.rs` file, so it loses its [`pub mod signature`](https://briansmith.org/rustdoc/src/ring/lib.rs.html#111) export. Ideally, since this `ring` functions as a dependency rather than a changeable source file, we would not skelefy it.

LukeMathWalker:
Now I got it 😁 
The second option you put forward seems to be the most appropriate. I don't fully understand the con you mentioned though:

> Forces users to list all directories in a dependency as a patch, even if they don't need it for the project themselves. For example, someone might need only a couple of Rocket's crates, but they would have to add all of Rocket's sub directories as patches, even though they don't need them.

Can you elaborate? Once again, an example would be ideal.

prestontw:
> I still don't fully understand the con you mentioned though:
> 
> > Forces users to list all directories in a dependency as a patch, even if they don't need it for the project themselves. For example, someone might need only a couple of Rocket's crates, but they would have to add all of Rocket's sub directories as patches, even though they don't need them.
> 
> Can you elaborate?

I'll describe this with another example first, then create a minimal repo (just in case it's clear enough that a minimal repo isn't necessary).

We patch `Pear`, I'm not sure what version since its `Cargo.toml` is especially bare. Our dependency is on the [`Pear/lib` member](https://github.com/SergioBenitez/Pear/blob/c674617b6dfb008831de7032cc771352c0027a9c/Cargo.toml#L3). However, a file in that member has a dependency on another member, specifically, `codegen` for its [`switch` proc macro](https://github.com/SergioBenitez/Pear/blob/c674617b6dfb008831de7032cc771352c0027a9c/codegen/src/lib.rs#L397). Then when we skelefy `codegen`'s `lib.rs`, we obliterate this `switch` declaration:
```
#34 52.67 error[E0432]: unresolved import `switch`
#34 52.67  --> rust-patches/Pear/lib/src/parsers.rs:1:49
#34 52.67   |
#34 52.67 1 | use {Result, Input, Length, Expected, ParseErr, switch};
#34 52.67   |                                                 ^^^^^^ no `switch` in the root
#34 52.67 
#34 52.67 error: cannot determine resolution for the macro `switch`
#34 52.67    --> rust-patches/Pear/lib/src/parsers.rs:296:9
#34 52.67     |
#34 52.67 296 |         switch! { [collection; input]
#34 52.67     |         ^^^^^^
#34 52.67     |
#34 52.67     = note: import resolution is stuck, try simplifying macro imports
#34 52.67 
#34 53.10 error: aborting due to 2 previous errors
```

To try to rephrase
> Forces users to list all directories in a dependency as a patch, even if they don't need it for the project themselves.

I, the user, now have to discover the internal, implicit dependencies between the members of the entire `Pear` crate, even though I only depend on a specific member, through compiler error messages. Another way of phrasing this: as the user of a dependency, I now have to list the transitive dependencies within a rust-patch in my `Cargo.toml` even though I don't use them.

LukeMathWalker:
> Then when we skelefy `codegen`'s `lib.rs`, we obliterate this `switch` declaration:

I think this is the part that confuses me - why is `codegen` in your local filesystem if you are just patching `Pear/lib`?
I'd expect `Pear/lib` to pull crates.io version of `codegen`.

LukeMathWalker:
Or are we working under the assumption that the patched version of `Pear/lib` brings the whole workspace/large parts of it due to path dependencies?

prestontw:
> Or are we working under the assumption that the patched version of `Pear/lib` brings the whole workspace/large parts of it due to path dependencies?

This is correct. At least for the example, `Pear/lib` depends on `codegen` locally: https://github.com/SergioBenitez/Pear/blob/master/lib/Cargo.toml#L11

(I realized that one of my earlier points isn't as painful since we can look in `Cargo.toml`s to add `rust-patches` rather than adding based strictly on compiler errors, but it still leaks internal dependencies to users.)

prestontw:
> This is correct. At least for the example, `Pear/lib` depends on `codegen` locally: https://github.com/SergioBenitez/Pear/blob/master/lib/Cargo.toml#L11

I've been looking into recreating the issue in another repo, and it looks like we "patched" the dependencies by copying their entire `git` repos instead of de-`tar`ing the vendored tarballs (from using `cargo-local-registry`). So, this seems pretty unique. It's definitely possible for people to copy a git repo instead of manipulating the existing vendored tarball, so it might pop up in the future?

Separately, I'm going to look if this is an issue with using `cargo vendor` instead of `cargo-local-registry`.

prestontw:
In trying to recreate the issue, I've come across a separate issue using `cargo chef` with dependencies vendored with `cargo vendor`. I've made a minimal repo at https://github.com/prestontw/cargo-chef-vendored-repo



> The second option you put forward seems to be the most appropriate.

After investigating, this seems the most appropriate to me too---I don't think it's worth trying to approximate the patch directory based off of a weird historical vendor. If it makes sense to you, I'll revert this PR to a point where it was using the second option, then potentially raise another PR to address the issue in https://github.com/prestontw/cargo-chef-vendored-repo. How does that sound?

LukeMathWalker:
Sounds good to me!
I wonder if you can determine the local path dependencies using `Cargo.lock`, therefore fixing the issue of transitive "patched" dependencies in one go.

prestontw:
> I wonder if you can determine the local path dependencies using `Cargo.lock`, therefore fixing the issue of transitive "patched" dependencies in one go.

How do you imagine this working? I am checking `Cargo.lock`, and the only difference I can tell between patched dependencies and vendored dependencies is patched dependencies don't have a `source` specified.

prestontw:
> I wonder if you can determine the local path dependencies using `Cargo.lock`, therefore fixing the issue of transitive "patched" dependencies in one go.

We could try to determine local path dependencies by using patched crates' `Cargo.toml`'s. We would examine their dependencies for explicit `path`s and add those crates to the list of crates we shouldn't skelefy? I'm also wondering if I should just list out these transitive dependencies as patches (at least in my case).

LukeMathWalker:
You are right unfortunately - the lockfile is not very helpful in this case. 
Traversing the manifests seems a good alternative option.

prestontw:
> Traversing the manifests seems a good alternative option.

Do you think that the complexity of traversing manifests is worth it? The alternative is adding more directories to the patches section of a `Cargo.toml`, which is a little strange to encounter. But I'm not sure if this is a common enough feature to add the complexity. I can also try doing this traversal and seeing if it makes the code harder to follow---what do you think?

LukeMathWalker:
> > Traversing the manifests seems a good alternative option.
> 
> Do you think that the complexity of traversing manifests is worth it? The alternative is adding more directories to the patches section of a `Cargo.toml`, which is a little strange to encounter. But I'm not sure if this is a common enough issue to add the complexity. I can also try doing this traversal and seeing if it is actually complex---what do you think?

I think that tools like `cargo-chef`, if they want to be adopted and trusted, should strive to keep the simplest interface with the minimal possible number of downsides.
So I definitely consider it worth it.

Another possible avenue: `cargo +nightly build --unit-graph -Z unstable-options` returns the source path for each dependency in the tree, although it requires nightly.

alsuren:
> Another possible avenue: `cargo +nightly build --unit-graph -Z unstable-options` returns the source path for each dependency in the tree, although it requires nightly.

It is possible to do the same with the output of `cargo metadata` on stable. I hacked something up the other day to help cargo watch find relative dependencies. Basically traverse the dependency graph, and skip anything that is a remote dependency by filtering out CARGO_HOME.

https://github.com/watchexec/cargo-watch/pull/192/files#diff-2b01ecd6b65bb66a3751d23fd19e3fdb6f2764fc466b7537438d04719f4257a3R44

This take a couple of seconds, but it *works*, and handles patches and relative dependencies without having to parse Cargo.toml or .cargo/config etc. The risk is that the cargo metadata command might hunt for cargo configs that are outside your docker context root, so your recipe JSON might change based on what someone has in their ~/.cargo/config if you're really unlucky? Maybe that's not a problem for us?

@prestontw I haven't looked at your code yet (I'm writing this on my phone) so I might be missing the point entirely. Do you think that this approach might solve the problem for you?

prestontw:
> Do you think that this approach might solve the problem for you?

Hmm, interesting! So in my particular case, rather than filtering based on whether the package is in `CARGO_HOME`, I would filter based on if they are in the patch directory? I might not be understanding fully, but I'm not sure if we can rely on there being a "patch directory"---this sounds like a way to implement approach 3 in https://github.com/LukeMathWalker/cargo-chef/pull/70#issuecomment-860817001. Maybe this gives us more fine-grained directory analysis, so rather than trying to infer whether they are in the patch directory, we can tell that we are a sibling of something that is patched?

Looking at the manifests' and pulling more directories in based on explicit paths sounds the most robust to me, but I'm curious what other people think. @alsuren am I understanding what you're suggesting?

LukeMathWalker:
Closing as stale.

PR #71, opened 2021-06-17
Build one Docker image per Rust version and tag with cargo-chef version number

### LINKED ISSUE #58: Versioned docker images
This is a just a feature request to ask: would it be possible to have tagged versions in docker registry? This would help with caching the container since "latest" may always get downloaded

### LINKED ISSUE #63: pre-compiled binary
Hi.

I would love to see cargo-chef had pre-compiled binary on the release page.
I tend to avoid invoking the `cargo install` command in CI workflow and downloading the pre-compiled binary directly.
Resulting in just ~14 minutes, compared to ~30 minutes.

So instead of `cargo install trunk`, I do:

```bash
      # install trunk
      wget -qO- https://github.com/thedodd/trunk/releases/download/v0.10.0/trunk-x86_64-unknown-linux-gnu.tar.gz | tar -xzf-
      chmod +x trunk
      cp trunk ~/.cargo/bin/
      trunk --version
```

I prefer this approach than using `lukemathwalker/cargo-chef` as the base image. So I have more fined grain control over the image.

jmagnusson:
Pulls current Rust docker image tags from [the official GitHub repo](https://github.com/docker-library/official-images/blob/master/library/rust).

Image tag format looks like this: `{cargo chef version}-rust-{dockerhub rust image tag}`. 

For non-alpha/beta versions of cargo-chef there's also a `latest-rust-{dockerhub rust image tag}` equivalent tag pushed, and a `latest` tag like before (which uses `rust:latest`).

E.g. for a non-alpha/beta version some tags pushed might be:

- `latest`
- `0.1.22-rust-1`
- `latest-rust-1`
- `latest-rust-1.52-buster`
- `0.1.22-rust-1.52.1-alpine3.12`
- `latest-rust-1.52.1-alpine3.12`
- etc...

For alpha/beta versions:

- `0.1.22-alpha.1-rust-1`
- `0.1.22-alpha.1-rust-1.52.1-alpine3.12`
- etc...

I set up a DockerHub repo while testing my changes. So examples of built images can be found here: https://hub.docker.com/repository/docker/jacobsvante/cargo-chef/tags. They were built by this GitHub Actions run: https://github.com/jmagnusson/cargo-chef/actions/runs/946755916

Fixes #58

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,12 +1,16 @@
 name: Build Docker images
 on:
   push:
-    tags:
-      # Do not build on alpha tags!
-      - v[0-9]+.[0-9]+.[0-9]+
-      # NOTE: The job relies on above format being used
+    branches: [main, master]
   schedule:
     - cron: '42 7 * * *' # run at 7:42 UTC (morning) every day
+
+env:
+  RUST_IMAGE_TAG: latest
+  DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+  DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+  DOCKER_REPO: ${{ secrets.DOCKERHUB_USERNAME }}/cargo-chef
+
 jobs:
   rust_image_tag_matrix:
     name: Generate Rust Docker image tag matrix
@@ -34,10 +38,6 @@ jobs:
       fail-fast: false
       matrix:
         rust_image_tag: ${{fromJSON(needs.rust_image_tag_matrix.outputs.matrix)}}
-    env:
-      RUST_IMAGE_TAG: latest
-      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
-      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
     steps:
       -
         name: Checkout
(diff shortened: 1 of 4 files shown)


jmagnusson:
I would recommend that the Github Actions matrix be dynamically populated with current Rust tags in the future. Having Rust version manually specified gets outdated pretty fast.

This can be done with fromJSON: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#fromjson.

jmagnusson:
New commit pushed which changes the behavior to always run Docker build/push for the main/master branch. `cargo-chef` version is picked up from Cargo.toml, meaning that even alpha/beta versions will be pushed to Docker Hub (although without latest tags).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-chef"
-version = "0.1.22-alpha.0"
```
LukeMathWalker, line 3: We are not cutting a release for this PR (given that it does not bring functional changes to the binary).

.github/workflows/docker.yml
```diff
@@ -1,42 +1,96 @@
 name: Build Docker images
 on:
   push:
-    tags:
-      # Do not build on alpha tags!
-      - v[0-9]+.[0-9]+.[0-9]+
+    branches: [main, master]
```
LukeMathWalker, line 4: I'd revert this back to building on tags - tags + building on a schedule to get new Rust releases is enough.

LukeMathWalker:
Massive thanks for having worked on this ❤️ 
The auto-generated Rust matrix version is :chef-kiss:

I have left just two tiny comments - once those are addressed we can merge 🚀

Cargo.toml
jmagnusson: Agreed! I left the version change by mistake while testing the jobs... Fixing now!

.github/workflows/docker.yml
jmagnusson: The declared schedule wasn't triggering for me. So I tested out this theory that it had something to do with tags being looked at in the workflow. For some reason the schedule started working after I made this change.

The version is now parsed from the Cargo.toml file so the filter that was specified previously is no longer needed IMO. So I'll change to just building on new tags. Ok?

jmagnusson:
NP! Thanks for this awesome project 👍

.github/workflows/docker.yml
LukeMathWalker: I'd still keep the filter - we don't want to build on alpha tags, which might still happen from time to time.

jmagnusson: What if someone wants access to a feature that's being worked on / is about to be released? Then they could reference `0.1.337-beta.1-rust-1`.

Note that latest tags will not be built for alpha/beta/rc builds. You will have to reference the alpha/beta tag explicitly.

LukeMathWalker: Ok, you convinced me, let's leave it 😁

jmagnusson: Yay! 😅

jmagnusson: @LukeMathWalker So time to merge?

PR #72, opened 2021-06-30
Add a --check flag

snoyberg:
I'm working on a CI setup where I want to be able to run both `cargo clippy` and `cargo build`. Without the `--check` flag, we need to rebuild the dependencies for checking each time. With this addition, `cargo clippy` is able to leverage the cache.

Thanks for the wonderful tool, it's really sped up our CI!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,6 +60,9 @@ pub struct Cook {
     /// Build in release mode.
     #[clap(long)]
     release: bool,
+    /// Check instead of build.
+    #[clap(long)]
+    check: bool,
     /// Build for the target triple.
     #[clap(long)]
     target: Option<String>,
@@ -112,6 +115,7 @@ fn _main() -> Result<(), anyhow::Error> {
         Command::Cook(Cook {
             recipe_path,
             release,
+            check,
             target,
             no_default_features,
             features,
@@ -178,6 +182,7 @@ fn _main() -> Result<(), anyhow::Error> {
             recipe
                 .cook(CookArgs {
                     profile,
+                    check,
                     default_features,
                     features,
                     target,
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
Hi @snoyberg!

Can you tell me a bit more about your usecase?
I am a bit surprised to hear you are using `cargo-chef` for your CI pipeline - wouldn't `sccache` or just plain-old incremental compilation (by caching the right folders) be a better match?

snoyberg:
Sure, happy to share. I regularly perform builds within Docker containers in our CI setups. We _could_ use CI caching to speed this up, and do that in plenty of other cases. But performing the build within a Docker image can provide us with a more reliable CI "caching" strategy, by pinning down exactly which tag of the base build image we're using. Specifically in this case, I'm performing a statically linked MUSL-based build, and knowing that we have all of the base system _and_ Rust dependencies fully configured is nice.

I'd be happy to extract the relevant files from our project setup if you'd like to see. I know not everyone is in favor of the concept of performing builds this way. But I actually had thought that this was one of the primary use cases of cargo-chef. If our Docker registry had better `--cache-from` support, it would have been a simple multistage Docker use case.

src/main.rs
```diff
@@ -60,6 +60,9 @@ pub struct Cook {
     /// Build in release mode.
     #[clap(long)]
     release: bool,
+    /// Check instead of build.
```
LukeMathWalker, line 63:
```suggestion
    /// Run `cargo check` instead of `cargo build`. Primarily useful for speeding up your CI pipeline.
```

LukeMathWalker:
That's enough for me to move forward 😁 Although, if you want to share a minimal example of your setup, I am sure others will benefit! (And I am curious as well, of course)

snoyberg:
Awesome, thanks for the merge and release! I've put together a GitLab repository demonstrating the workflow I'm using on some projects, together with a README explaining it:

https://gitlab.com/snoyberg/docker-chef-ci-example

PR #73, opened 2021-07-12
Add flag for cooking #[no_std] crates

mkroening:
This adds support for using `cargo-chef` with `#[no_std]` crates.

When specifying `--no-std`, an appropriate `#[no_std]` template will be used for generating the skeleton.

On `proc-macro` crates and build scripts, this will be ignored. This allows `#[no_std]` workspaces to be built using `--workspaces`, even when they include `proc-macro` crates. When building workspaces which mix `std` and `#[no_std]` crates, `--no-std` has to be provided on a per crate basis, as there is no easy way to determine the `std`-ness of a crate, as far as I know.

What do you think?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,7 @@ dependencies = [
  "globwalk",
  "log",
  "pathdiff",
+ "predicates",
  "serde",
  "serde_json",
  "toml",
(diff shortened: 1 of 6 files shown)


LukeMathWalker:
Looking good!
Sorry it took me so long to review this 😅 

Would you mind adding some tests for the new functionality?

mkroening:
No worries!

I added no_std tests, asserting the content of the source files.

mkroening:
Rebased on latest `main`.

LukeMathWalker:
The failing test is due to an issue with ordering - I'll be fixing that in a separate PR.
Thanks for the contribution!

PR #76, opened 2021-08-10
fix version masking

conradludgate:
I noticed when testing that my recipe.json was being rebuilt when I updated my package version number

### DIFF
--- a/src/skeleton.rs
+++ b/src/skeleton.rs
@@ -45,10 +45,10 @@ impl Skeleton {
                     let mut intermediate = toml::Value::try_from(parsed)?;
 
                     // ignore package.version for recipe
-                    *intermediate
+                    intermediate
                         .get_mut("package")
                         .and_then(|v| v.get_mut("version"))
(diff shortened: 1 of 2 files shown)


conradludgate:
Seems to also be included in the Cargo.lock file, and there's currently no logic to mask it there. Should that also be done?

LukeMathWalker:
I am not 100% sure on what behaviour you are changing - isn't the old `main` version equivalent to your branch version?

conradludgate:
So the code I replaced makes it seem that the Cargo.toml `package.version` value should be replaced with `0.1.0`, but when I run `cargo chef prepare`, it is not getting replaced and the resulting `recipe.json` file is different (meaning my cache is invalidated and needs a rebuild of dependencies)

LukeMathWalker:
Understood - would you mind adding a test, so that we do not regress on this functionality going forward?

LukeMathWalker:
Almost there - just run `cargo fmt` and then we can merge 😁

conradludgate:
> Almost there - just run `cargo fmt` and then we can merge grin

There's always something :facepalm:

Issue #77, opened 2021-08-15
Fail to build due to change in `clap`


johnmave126:
Although the dependency of `clap` is specified as `3.0.0-beta.2`, it turns out that when performing `cargo install cargo-chef`, cargo always pulls the newest version matching semvar, ignoring the beta part. So at this moment, cargo will pull `3.0.0-beta.4` version of `clap` to build `cargo-chef`.

Due to change in https://github.com/clap-rs/clap/pull/2686
https://github.com/LukeMathWalker/cargo-chef/blob/b331984848ee760c9099d5a95fd700cbe7358ce1/src/main.rs#L76
fails to compile since `value_delimiter` now accepts `char` instead of `&str`.

I will do a fork and submit a pr for this.

BeneSim:
i did comment the PR but I'd also like to add that as a workaround you could simply use `cargo install cargo-chef --locked` at the moment to use the `Cargo.lock` file which does work. You won't get any potential security updates whatsoever but at least you'd have a working CI/CD pipeline again until the PR from @johnmave126 gets merged.

LukeMathWalker:
My mistake here - I should have pinned `clap`'s version using `=`, preventing `cargo` from pulling in the latest semver-compatible version.

PR #78, opened 2021-08-15
Bump clap dependency to 3.0.0-beta.4

johnmave126:
Fix https://github.com/LukeMathWalker/cargo-chef/issues/77

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,25 +4,26 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.14"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b476ce7103678b0c6d3d395dbbae31d48ff910bd28be979ba5d48c6351131d0d"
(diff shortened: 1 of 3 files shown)


BeneSim:
I guess you'd also need to adjust at least the https://github.com/LukeMathWalker/cargo-chef/blob/main/README.md since it references 1.53 a couple of times (might as well need to change some Dockerfiles etc.) and the new beta.4 is incompatible with Rust 1.53 (due to the usage of macros in attributes which became stable in 1.54).

To be honest I'd prefer if cargo-chef moved away from using beta crates since it can be very annoying. I guess clap-3.0.0 could be easily replaced with structopt since it's basically the same nowadays. 

I guess most of the people use cargo-chef in their CI/CD, and that should be as stable as it gets, failing CI pipelines because breaking changes in a beta crate are just nasty. Especially since they propagate into many versions, i.e. we're currently using cargo-chef 0.1.11 and are affected by this too.

Btw. simply using comparison version matching for `clap-3.0.0.beta.2` won't work since `clap-3.0.0.beta.2` itself doesn't use comparison match for `clap-derive ....` ....

LukeMathWalker:
Let's avoid bumping up the minimum required version to 1.54 - I've fixed the issue in #79.

PR #80, opened 2021-08-16
Suggest using --locked in README

Kinrany:
Since cargo-chef is intended for automated setups, it's better for reproducibility to use the binary's lock file during installation.

There's a relevant comment about `--locked` in `cargo install` docs: https://doc.rust-lang.org/cargo/commands/cargo-install.html


LukeMathWalker:
Thanks!

Issue #81, opened 2021-08-17
version invalidation


conradludgate:
Related to #76, not only does the `Cargo.toml` file contain the version of the package, but so does [`Cargo.lock`](https://github.com/LukeMathWalker/cargo-chef/blob/5fb7e970c6dbf8a879e0293285eba4860682c5b0/Cargo.lock#L82-L84). This means that despite only a version change, the cache will still be invalidated.

Currently there's no processing on the Cargo.lock file like there is on Cargo.toml. I see two possibilities,

1. Make use of the fact that Cargo.lock should be generated automatically, assuming the formatting.
```toml
[[package]]
name = "cargo-chef"
version = "0.1.24-alpha.0"
```
We know the version from the previous Cargo.toml formatting stage and we know the package name, so we can do a simple string find and replace

2. Parse the toml, loop through the packages til we find the one with our package name, then mutate the version. Finally reconverting the value back into a string.

Option 2 is probably more correct

LukeMathWalker:
Option 2. sounds good - let's work in #82.

LukeMathWalker:
#82 has been merged and released as part of `0.1.24`.
There is only one TODO left, mostly as an additional check: verifying that our substitution logic works on all versions of the lockfile.
Ideally we want a test for version `1` and `2` in addition to the one you added for version `3`. Fancy giving it a go?

conradludgate:
There's only one type of lockfile this won't work on: https://github.com/rust-lang/cargo/blob/dd8fa9183d/tests/testsuite/lockfile_compat.rs#L37-L49 (using `root` for the package) - admittedly, this is a very old format and was [removed 5 years ago](https://github.com/rust-lang/cargo/commit/5e644df1fffc5726b2bf4ce8e29990602208ac51)

However, if we want to support this, it should be pretty straightforward.

I've tested all the other formats and they all work and my layers are cached between versions :)

LukeMathWalker:
5 years ago is plenty of time - no need to go that far away unless somebody has a compelling usecase for it.
I'd say we can close this then - thanks for working on it!

PR #82, opened 2021-08-17
feat: mask the version in cargo.lock

conradludgate:
for #81


src/skeleton.rs
```diff
@@ -32,6 +32,7 @@ impl Skeleton {
             .build()
             .context("Failed to scan the files in the current directory.")?;
         let mut manifests = vec![];
+        let mut package_names = vec![];
```
LukeMathWalker, line 35: Perhaps `local_package_names` would be a better name?

LukeMathWalker:
Thanks!

PR #87, opened 2021-08-25
Fix published versions not being pushed to Docker Hub

### LINKED ISSUE #86: Only pre-release versions are pushed to Docker Hub?
Hi, thanks for this awesome project!

It seems that all images on docker hub have `-alpha.0` suffix in their version part (e.g. `0.1.26-alpha.0-rust-latest`, `0.1.25-alpha.0-rust-latest`).
https://hub.docker.com/r/lukemathwalker/cargo-chef/tags

I think these are pre-release versions, but I can't find any version tagged without this suffix (e.g. `0.1.26-rust-latest`, `0.1.25-rust-latest`).
Is this expected behavior?

jmagnusson:
Fixes #86

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,6 +2,7 @@ name: Build Docker images
 on:
   push:
     branches: [main, master]
+    tags: '*'
   schedule:
     - cron: '42 7 * * *' # run at 7:42 UTC (morning) every day
 


LukeMathWalker:
This does not fully solve the problem - when the build is triggered according to the cron schedule, it will still pick up the alpha version.
We need the cron build to pick up the newer version of the compiler.
I have implemented a fix in #89, so I'll close this one.

PR #88, opened 2021-08-28
Fix Rust CI trigger

sksat:
SSIA

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -1,6 +1,20 @@
 name: Rust
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
Thanks for opening a PR!
Unfortunately, I won't be going forward with a merge: although this technically reduces the scope of changes triggering a CI run to the bare minimum required, it adds additional maintenance overhead. If tomorrow we start embedding our README.md file for module docs, add an examples folder or change the layout of the source code (e.g. convert to a workspace) we need to remember to go and change the selectors in the CI pipeline. We will probably forget and stuff will fall through the cracks.
In practice, the excluded files are rarely changed so it is easier to leave the trigger as is.

PR #96, opened 2021-09-26
fix alpine instructions

devdoshi:
The alpine build instructions in the readme will fail if copied because some arguments were being passed at the wrong time. This change makes it work


LukeMathWalker (approved):
Thanks!

PR #118, opened 2022-02-27
Update clap from v3.0 beta 2 to v3.1

LovecraftianHorror:
Nothing too exciting, just updates clap to the most recent version.

This change is actually because I noticed a slight inconsistency where `cargo` would accept `--features=` as default features while `cargo-chef` rejected it as missing a value. Looking into this further, I noticed that `clap` changed this behavior to match `cargo` in one of the more recent releases, so this PR fixes this minor CLI inconsistency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,6 @@ dependencies = [
  "atty",
  "cargo-manifest",
  "clap",
- "clap_derive",
  "env_logger",
  "expect-test",
  "fs-err",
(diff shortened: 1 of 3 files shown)


LovecraftianHorror:
Looks like the `clippy` failure is from a new or updated lint unrelated to these changes. I can fix it if desired

LukeMathWalker:
It's alright, I can take care of the lint. Thanks for the PR!

PR #119, opened 2022-03-05
Publish linux musl binaries on release too

LovecraftianHorror:
This adds publishing Linux musl based binaries to the releases. The changes are mostly a blend of the current implementation and [`cargo-hack`'s upload-assets job](https://github.com/taiki-e/cargo-hack/blob/main/.github/workflows/release.yml#L47-L82)

This is useful for when you need a musl build and the pre-build images don't quite cut it like building an alpine image with a nightly toolchain for instance


LukeMathWalker (approved):
Thank you!

Issue #74, opened 2021-08-05
`cargo-chef` isn't local registry aware


prestontw:
Link to minimal repo: https://github.com/prestontw/cargo-chef-vendored-repo

I've vendored the dependencies of a small Rocket app using `cargo vendor`. I've then updated `.cargo/config.toml` so that `cargo build` uses the local dependencies. Trying to run `cargo chef` errors with checksum changes (since cargo chef has replaced all of the `lib.rs` files in the dependencies with empty files?).

I'm thinking if we make `cargo chef` aware of `.cargo/config.toml`, we will know not to skelefy these vendored files.

Aside: using `cargo-local-registry` is fine since it stores dependencies in zipped files, so `cargo chef` doesn't skelefy them.

LukeMathWalker:
`.cargo/config.toml` is already part of the recipe file, so it should be possible to make this work without any changes to the CLI parameters.

PR #120, opened 2022-03-21
Ptw/issue 74 Do not skelefy vendored directory

prestontw:
This is a quick stab at addressing https://github.com/LukeMathWalker/cargo-chef/issues/74.

There are two design questions: 
1. whether we filter these vendored `Cargo.toml` files from our list of manifests, or include them in the skeleton and then filter them out when we are `cooking`, and
2. How much do we want to hardcode to `cargo vendor`'s behavior.

Regarding 1., I prefer to filter these vendored files from our list of manifests. This makes testing easier (than having to render the recipe then checking directory outputs), keeps the manifests in the skeleton we serialize focused on what which projects we need to skelefy, and is robust against vendored files changing: if a vendored file changes the Docker sha of the step will change so it doesn't matter that the recipe.json file doesn't change---the recipe file will be recomputed by Docker anyway since one of its dependencies in the Dockerfile changed. I am happy to listen to feedback and preferences if this is not the direction the project wants to go.

Regarding 2., `cargo vendor`'s configuration currently looks like: 
```
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
```
while `cargo-local-registry` is something like:
```
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
local-registry = "vendor"
```
(cargo local registry downloads sources as tarballs, so cargo chef happily ignores them)

I think that going through `crates-io` to the other field is more robust to different project structures, though I was definitely tempted to go straight to `vendored-sources`.

### DIFF
--- a/src/skeleton/read.rs
+++ b/src/skeleton/read.rs
@@ -38,7 +38,7 @@ pub(super) fn manifests<P: AsRef<Path>>(
 ) -> Result<Vec<ParsedManifest>, anyhow::Error> {
     let vendored_path = vendored_directory(config_contents);
     let builder = if let Some(path) = vendored_path {
-        let exclude_vendored_sources = "!".to_string() + &path;
+        let exclude_vendored_sources = format!("!{}", path);
         GlobWalkerBuilder::from_patterns(&base_path, &["/**/Cargo.toml", &exclude_vendored_sources])
     } else {
         GlobWalkerBuilder::new(&base_path, "/**/Cargo.toml")
--- a/src/skeleton/read.rs
+++ b/src/skeleton/read.rs
@@ -21,10 +21,29 @@ pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, an
     }
 }
 
+fn vendored_directory(config_contents: Option<&str>) -> Option<String> {
+    config_contents
+        .and_then(|contents| toml::Value::try_from(contents).ok())
+        .and_then(|contents| {
+            contents
+                .get("source.crates-io")
+                .map(|value| value.to_owned())
+        })
+        .and_then(|value| value.as_str().map(|s| s.to_owned()))
+}
+
 pub(super) fn manifests<P: AsRef<Path>>(
     base_path: &P,
+    config_contents: Option<&str>,
 ) -> Result<Vec<ParsedManifest>, anyhow::Error> {
-    let walker = GlobWalkerBuilder::new(&base_path, "/**/Cargo.toml")
+    let vendored_path = vendored_directory(config_contents);
+    let builder = if let Some(path) = vendored_path {
+        let exclude_vendored_sources = "!".to_string() + &path;
+        GlobWalkerBuilder::from_patterns(&base_path, &["/**/Cargo.toml", &exclude_vendored_sources])
+    } else {
+        GlobWalkerBuilder::new(&base_path, "/**/Cargo.toml")
+    };
+    let walker = builder
         .build()
         .context("Failed to scan the files in the current directory.")?;
 
--- a/src/skeleton/read.rs
+++ b/src/skeleton/read.rs
@@ -59,7 +59,6 @@ pub(super) fn manifests<P: AsRef<Path>>(
                 parsed.complete_from_path(&absolute_path)?;
 
                 let mut intermediate = toml::Value::try_from(parsed)?;
-                println!("{:?}", intermediate);
 
                 // Specifically, toml gives no guarantees to the ordering of the auto binaries
                 // in its results. We will manually sort these to ensure that the output
--- a/src/skeleton/read.rs
+++ b/src/skeleton/read.rs
@@ -22,14 +22,14 @@ pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, an
 }
 
 fn vendored_directory(config_contents: Option<&str>) -> Option<String> {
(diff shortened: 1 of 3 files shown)


prestontw:
I verified that this fixes the issue in the [minimal repo](https://github.com/prestontw/cargo-chef-vendored-repo). Before the [fix](https://github.com/prestontw/cargo-chef-vendored-repo/commit/ec7ca8157f16986e540dc145371d4fd1fdeb6369):
```
 => ERROR [cacher 9/9] RUN cargo chef cook --offline --recipe-path recipe.json                                                                                                                               0.5s
------
 > [cacher 9/9] RUN cargo chef cook --offline --recipe-path recipe.json:
#20 0.463 error: failed to select a version for the requirement `rocket = "^0.5.0-rc.1"`
#20 0.463 candidate versions found which didn't match: 0.0.1
#20 0.463 location searched: directory source `/opt/hellow/vendor` (which is replacing registry `https://github.com/rust-lang/crates.io-index`)
#20 0.463 required by package `hellow-rocket v0.0.1 (/opt/hellow)`
#20 0.463 perhaps a crate was updated and forgotten to be re-vendored?
#20 0.463 As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
#20 0.465 thread 'main' panicked at 'Exited with status code: 101', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.31/src/recipe.rs:145:27
#20 0.465 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
------
failed to solve: executor failed running [/bin/sh -c cargo chef cook --offline --recipe-path recipe.json]: exit code: 101
```
With installing this version:
```
 => CACHED [cacher 9/9] RUN cargo chef cook --offline --recipe-path recipe.json                                                                                                                              0.0s
 => CACHED [builder  8/10] COPY --from=cacher /opt/hellow/target target                                                                                                                                      0.0s
 => CACHED [builder  9/10] COPY --from=cacher /usr/local/cargo /usr/local/cargo                                                                                                                              0.0s
 => CACHED [builder 10/10] RUN cargo build --offline   && cp target/debug/hellow-rocket /opt/hellow/hellow-rocket                                                                                            0.0s
 => CACHED [runtime 2/2] COPY --from=builder /opt/hellow/hellow-rocket /usr/local/bin/hellow-rocket                                                                                                          0.0s
 => exporting to image                                                                                                                                                                                       0.0s
 => => exporting layers                                                                                                                                                                                      0.0s
 => => writing image sha256:80d46587ab615528988952282ffe4bf3a28bdaf31896250068f818dd1d9d966d                                                                                                                 0.0s
 => => naming to docker.io/library/cargo-chef-vendored-repo_hellow-rocket 
```

prestontw:
> Regarding 2., `cargo vendor`'s configuration currently looks like:
> 
> ```
> [source.crates-io]
> replace-with = "vendored-sources"
> 
> [source.vendored-sources]
> directory = "vendor"
> ```

[Changing the directory into which I vendored](https://github.com/prestontw/cargo-chef-vendored-repo/commit/f9f973596d273fb269c9cc69928483c258d48147), I see that `vendored-sources` is [actually constant](https://github.com/prestontw/cargo-chef-vendored-repo/blob/f9f973596d273fb269c9cc69928483c258d48147/.cargo/config.toml#L4-L5):
```
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "local-registry"
```
So, regarding 2. above, we could hard code it. Pro: simpler to understand. Con: maybe not the most robust to future changes?

LukeMathWalker:
Hey @prestontw, thanks for picking this up 😁 

What do you mean by

> [Changing the directory into which I vendored](https://github.com/prestontw/cargo-chef-vendored-repo/commit/f9f973596d273fb269c9cc69928483c258d48147), I see that vendored-sources is [actually constant](https://github.com/prestontw/cargo-chef-vendored-repo/blob/f9f973596d273fb269c9cc69928483c258d48147/.cargo/config.toml#L4-L5):

?

src/skeleton/read.rs
```diff
@@ -21,10 +21,29 @@ pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, an
...
+}
+
 pub(super) fn manifests<P: AsRef<Path>>(
     base_path: &P,
+    config_contents: Option<&str>,
 ) -> Result<Vec<ParsedManifest>, anyhow::Error> {
-    let walker = GlobWalkerBuilder::new(&base_path, "/**/Cargo.toml")
+    let vendored_path = vendored_directory(config_contents);
+    let builder = if let Some(path) = vendored_path {
+        let exclude_vendored_sources = "!".to_string() + &path;
```
LukeMathWalker, line 41: Let's use `format!` here.

prestontw: Done as of [07ab6bd](https://github.com/LukeMathWalker/cargo-chef/pull/120/commits/07ab6bdf2c97fa07874a7dd38fc24d05e9a328f0)!

LukeMathWalker:
I understand now, thanks for explaining!
I'd still suggest to navigate through the chain in order to have an implementation that makes minimal assumptions about the behaviour of tooling related to vendoring.

LukeMathWalker (approved):

Issue #125, opened 2022-04-17
Feature request: automatically edit `Cargo.toml` for large workspaces


prestontw:
Hello! My workspace at work is _massive_. We build Docker images for individual components from this monorepo. Rather than copying over the entire monorepo for each individual component, we copy over our main monolithic backend directory or smaller individual services directories.

All of these `Dockerfile`s have 
```
RUN perl -0777 -i -pe 's/members = \[[^\]]+\]/members = ["individual-component"]/igs' Cargo.toml
```
because `cargo` will still compile other components even when specifying `--package` or `--bin`. 

I would like to request an enhancement: that cargo chef would handle this automatically. I have built a small repo that replicates this behavior: https://github.com/prestontw/cargo-chef-workspace-experiment. 

I could imagine that specifying `--bin`, `--package`, or something else (maybe `--only`?) would edit the `Cargo.toml` file in the event where the entire folder structure isn't copied over.

prestontw:
Looking at the code (and since this happens in the same stage as `prepare`), I would guess that this would be a flag to `Prepare` rather than `Cook`. (Wow! It seems like most flag work is for `Cook`. Interesting change of pace!)

prestontw:
Thinking about it some more, I think a `--member` argument might work nicely. It's distinct from `--package` to avoid conflating usage and more clearly (hopefully) conveys that it would edit the `members` field. I might take a quick stab at implementing this.

LukeMathWalker:
> because cargo will still compile other components even when specifying --package or --bin.

I am not sure this is true. I cloned your repository and ran `cargo b --bin backend`: `cargo` only built the `backend` crate. You can double-check this as well using the `--timings` flag:

![image](https://user-images.githubusercontent.com/20745048/163780534-0880c93d-d842-4add-9439-54a214910ccd.png)

prestontw:
> > because cargo will still compile other components even when specifying --package or --bin.
> 
> I am not sure this is true. I cloned your repository and ran `cargo b --bin backend`: `cargo` only built the `backend` crate. 

Ahh, nice. I should have been more precise. `cargo` will not compile `ci`, but it will check that its folder, workspace, and Cargo.toml structure is correct.

Replicating what you did (on commit `e0196d317e20cf1a8e80784d484f4903b40e5799` in the repro repo):
```
preston@Prestons-MBP cargo-chef-workspace-experiment % cargo b --bin backend
[...]

PR #126, opened 2022-04-18
Add optional member flag to `prepare`

prestontw:
If specified, and if the top-level cargo.toml
has a members list, replace the list with
only the specified member. 

This is an attempt to address https://github.com/LukeMathWalker/cargo-chef/issues/125. 
If you have feedback on naming or code structure (I feel a little weird placing this function in `mod.rs`),
please let me know! I am happy to make those changes.

I have verified that this works for https://github.com/prestontw/cargo-chef-workspace-experiment

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -58,7 +58,7 @@ pub struct Prepare {
     /// the whole workspace over. Cargo will complain that it can't
     /// find those members, so this replaces all of the members.
     #[clap(long)]
-    member: Option<String>,
+    bin: Option<String>,
 }
 
 #[derive(Parser)]
@@ -114,6 +114,12 @@ pub struct Cook {
     /// Cook using `#[no_std]` configuration  (does not affect `proc-macro` crates)
     #[clap(long)]
     no_std: bool,
+    /// Optional: the member in the workspace we wish to compile.
+    /// Note that this is useful in scenarios where we don't copy
+    /// the whole workspace over. Cargo will complain that it can't
+    /// find those members, so this replaces all of the members.
+    #[clap(long)]
+    bin: Option<String>,
 }
 
 fn _main() -> Result<(), anyhow::Error> {
@@ -143,6 +149,7 @@ fn _main() -> Result<(), anyhow::Error> {
             workspace,
             offline,
             no_std,
+            bin: _bin,
         }) => {
             if atty::is(atty::Stream::Stdout) {
                 eprintln!("WARNING stdout appears to be a terminal.");
@@ -211,12 +218,9 @@ fn _main() -> Result<(), anyhow::Error> {
                 })
                 .context("Failed to cook recipe.")?;
         }
-        Command::Prepare(Prepare {
-            recipe_path,
-            member,
-        }) => {
+        Command::Prepare(Prepare { recipe_path, bin }) => {
             let recipe =
-                Recipe::prepare(current_directory, member).context("Failed to compute recipe")?;
+                Recipe::prepare(current_directory, bin).context("Failed to compute recipe")?;
             let serialized =
                 serde_json::to_string(&recipe).context("Failed to serialize recipe.")?;
             fs::write(recipe_path, serialized).context("Failed to save recipe to 'recipe.json'")?;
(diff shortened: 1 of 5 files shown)


src/skeleton/mod.rs
```diff
@@ -29,9 +29,15 @@ pub(in crate::skeleton) struct ParsedManifest {
...
     /// Find all Cargo.toml files in `base_path` by traversing sub-directories recursively.
-    pub fn derive<P: AsRef<Path>>(base_path: P) -> Result<Self, anyhow::Error> {
+    pub fn derive<P: AsRef<Path>>(
+        base_path: P,
+        member: Option<String>,
+    ) -> Result<Self, anyhow::Error> {
         // Read relevant files from the filesystem
         let config_file = read::config(&base_path)?;
+        if let Some(member) = member {
+            replace_members_with_specific(&base_path, member)?;
```
prestontw, line 39: This makes `derive` produce a side effect (modifies the file system). We do this shortly after `derive` (in `main` after preparing the skeleton), but this seemed easier to me than trying to exclude the top-level Cargo.toml if it's both a workspace and `member` is specified.

src/main.rs
```diff
@@ -52,6 +52,13 @@ pub struct Prepare {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
+
+    /// Optional: the member in the workspace we wish to compile.
+    /// Note that this is useful in scenarios where we don't copy
+    /// the whole workspace over. Cargo will complain that it can't
+    /// find those members, so this replaces all of the members.
```
LukeMathWalker, line 59: I don't think we need to mention this here. Alternatively, we should change the language. Something along the lines of:

```
When --bin is specified, `cargo-chef` will ignore all members of the workspace that are not necessary to successfully compile the specific binary.
```

src/main.rs
```diff
@@ -52,6 +52,13 @@ pub struct Prepare {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
+
+    /// Optional: the member in the workspace we wish to compile.
```
LukeMathWalker, line 56:
```suggestion
    /// Optional: the name of the binary we wish to compile.
```

src/skeleton/mod.rs
LukeMathWalker: I am not a fan - we should manipulate the manifest in memory in `derive` and keep it side-effect free.
You should be able to do this easily by iterating over the `serialised_manifests` collection.

prestontw: That makes sense to me! I'm doing it in `manifests`, though, instead of `serialized_manifests`, to avoid serializing then deserializing.

prestontw:
Hey, @LukeMathWalker! Any other feedback? Is there anything I can do to help get this merged in?

LukeMathWalker:
I am on holiday for the next couple of weeks. I'll try to do a PR review spree when I am back :)

prestontw:
> I am on holiday for the next couple of weeks. I'll try to do a PR review spree when I am back :)

Thank you! Enjoy your time off! :D

src/main.rs
```diff
@@ -136,6 +145,7 @@ fn _main() -> Result<(), anyhow::Error> {
             workspace,
             offline,
             no_std,
+            bin: _bin,
```
LukeMathWalker, line 148: Shouldn't we pass this over in order to invoke `cargo build` with the `--bin` flag?

prestontw: Added in 438f86a! Thank you for the review!

src/recipe.rs
```diff
@@ -80,7 +81,8 @@ fn build_dependencies(args: &CookArgs) {
         package,
         workspace,
         offline,
-        ..
```
prestontw, line 83: I removed this so the compiler would help direct new contributors to where they should add command functionality if they add a new field, but I'm happy to revert this change if it doesn't match the project idioms.

LukeMathWalker (approved):

LukeMathWalker:
Thank you for all the work on this!

PR #127, opened 2022-04-20
Ptw/detect missing members automatically

prestontw:
A new, automatic approach to fulfilling https://github.com/LukeMathWalker/cargo-chef/issues/125

This really only affect Rust projects that make use of a `workspace`. To summarize, this will check the `members` in the workspace against the `Cargo.toml` files that we were able to find. If we don't find a corresponding `Cargo.toml` for a `member`, we remove it from our `members` list.

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -35,9 +35,6 @@ impl Skeleton {
     pub fn derive<P: AsRef<Path>>(base_path: P) -> Result<Self, anyhow::Error> {
         // Read relevant files from the filesystem
         let config_file = read::config(&base_path)?;
-        // if let Some(member) = member {
-        // replace_members_with_specific(&base_path, member)?;
-        // }
         let mut manifests = read::manifests(&base_path, config_file.as_deref())?;
         remove_missing_members(&manifests, &base_path)?;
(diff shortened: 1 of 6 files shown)


src/skeleton/mod.rs
```diff
@@ -300,3 +304,34 @@ fn serialize_manifests(manifests: Vec<ParsedManifest>) -> Result<Vec<Manifest>,
...
+    let contents = fs::read_to_string(&top_level_path)?;
+    let mut top_level: toml::Value = toml::from_str(&contents)?;
+    if let Some(members) = top_level
+        .get_mut("workspace")
+        .and_then(|workspace| workspace.get_mut("members"))
+    {
+        let members = members.as_array_mut().unwrap();
+        let found_members = manifests
+            .iter()
+            .filter_map(|manifest| manifest.relative_path.parent().map(|path| path.to_owned()))
```
prestontw, line 327: I'm a little worried if we run this from not the root directory. If this is a concern, I think that we can join both the members and the found members with the base path to be resilient to this? Maybe?

prestontw:
Closing this in favor of https://github.com/LukeMathWalker/cargo-chef/pull/126, as discussed in #125

PR #129, opened 2022-05-11
Add support for --profile flag

### LINKED ISSUE #109: Add a --profile flag to mirror cargo commands
We'd like to standardize on `--profile` in favor of `--release` vs nothing because I prefer the explicit flag.

slinkydeveloper:
Fixes #109. This PR adds support for [profile flag](https://doc.rust-lang.org/cargo/reference/profiles.html), mimicking the same interface of `cargo build`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use chef::{CookArgs, DefaultFeatures, OptimisationProfile, Recipe, TargetArgs};
 use clap::crate_version;
 use clap::Parser;
@@ -61,6 +61,9 @@ pub struct Cook {
     /// It defaults to "recipe.json".
     #[clap(long, default_value = "recipe.json")]
     recipe_path: PathBuf,
+    /// Build artifacts with the specified profile.
+    #[clap(long)]
+    profile: Option<String>,
     /// Build in release mode.
     #[clap(long)]
     release: bool,
@@ -121,6 +124,7 @@ fn _main() -> Result<(), anyhow::Error> {
     match command {
         Command::Cook(Cook {
             recipe_path,
+            profile,
             release,
             check,
             target,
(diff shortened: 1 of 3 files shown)


slinkydeveloper:
Tested this locally and unfortunately it doesn't work because apparently cargo-manifest doesn't parse custom profiles, hence the recipe won't contain my custom profile. Even `cargo_toml` doesn't look like it can parse custom profiles: https://docs.rs/cargo_toml/0.11.5/cargo_toml/struct.Manifest.html#fields

Perhaps cargo code itself should be used for that? https://github.com/rust-lang/cargo/blob/master/src/cargo/util/toml/mod.rs#L420

LukeMathWalker:
We use a fork of `cargo-manifest` - you may need to submit a PR [here](https://github.com/LukeMathWalker/cargo-manifest) first.

slinkydeveloper:
Yeah I saw it, problem is that neither your fork nor `cargo_toml` has support for that. Do you think it's worth fixing this issue in `cargo-manifest`, or rather `cargo_toml` where you can pull from it?

slinkydeveloper:
Hey @LukeMathWalker sorry for the delay, here is the PR to fix the bug in cargo-manifest: https://github.com/LukeMathWalker/cargo-manifest/pull/11

LukeMathWalker (approved):

tillrohrmann:
I think the PR is still missing something. When trying out the latest build with this feature, I couldn't run `cargo chef cook --profile custom-profile` because it failed with ` error: profile `custom-profile` is missing an `inherits` directive (`inherits` is required for all profiles except `dev` or `release`)`. Checking `recipe.json` it shows that the `inherits` directive is being filtered out.

LukeMathWalker:
That's a gap in `cargo-manifest`, I patched it in https://github.com/LukeMathWalker/cargo-manifest/pull/12
Can you run `cargo update -p cargo-manifest` and try again?

tillrohrmann:
Thanks for the quick patching. Will give it another try.

PR #133, opened 2022-06-06
fix: deal with bins that don't have `path`

kizzx2:
Implicitly Cargo could find them as `src/bin/{name}.rs`


LukeMathWalker (approved):

LukeMathWalker:
Thank you!

PR #134, opened 2022-06-06
feat: flag dont_mask_local_crate_versions

kizzx2:
I had to create this for my project which has a local version of some of the dependencies (e.g. I forked `rust-decimal` which is required by some other dependencies). I believe the logic behind masking local crate version is that local crates wouldn't be referenced by other dependencies but for my case I had to disable masking for it to build.

Comments / modifications welcome


LukeMathWalker:
Hi 
Can you provide a reproducible example that shows why this change is required?

It should fail on the current released version, pass on this one.

PR #144, opened 2022-07-18
ci: Fix failing builds

jacobsvante:
This error seems to have been introduced by the environment variable added in https://github.com/LukeMathWalker/cargo-chef/commit/f57e866a546a1cbddb08091381cf09aa50b65aaf.

Perhaps we can wait with extra speed until sparse index is in stable..?


LukeMathWalker:
IIRC, I introduced that environment variable because all builds were failing due to how slow it was to fetch the index. Let's give it a go again.

Issue #93, opened 2021-09-04
Docker build error: candidate versions found which didn't match: 0.0.1


orhun:
Hello, thanks for this crate. I'm using it for the Docker builds of my [git-cliff](https://github.com/orhun/git-cliff) project for a couple of months now. But recently I got an error in my CI workflow:

```
 Step 8/20 : RUN cargo chef cook --release --recipe-path recipe.json
 ---> Running in e89886c52ec8
    Updating crates.io index
error: failed to select a version for the requirement `git-cliff-core = "^0.2.5"`
candidate versions found which didn't match: 0.0.1
location searched: /app/git-cliff-core
required by package `git-cliff v0.0.1 (/app/git-cliff)`
thread 'main' panicked at 'Exited with status code: 101', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.30/src/recipe.rs:145:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
The command '/bin/sh -c cargo chef cook --release --recipe-path recipe.json' returned a non-zero code: 101
```

Dockerfile: https://github.com/orhun/git-cliff/blob/72cee9caa632498c55a7c5f1aac2657fbb445b90/Dockerfile

I have already seen #84 and although it seems like the same issue, I realized I can build it just fine with `cargo-chef v0.1.23` so I suspect this is happening because of the things changed between `v0.1.23`-`v0.1.24`.

I'm not sure how `cargo-chef` works internally so I cannot debug this further for now. Any ideas about how to fix this?

LukeMathWalker:
Thanks for reporting!
It should be fixed in `0.1.31`.

orhun:
It's building without errors now. Thanks!

sebastiendan:
Hi, I'm getting the same error now with `v0.1.41`. You can reproduce it by preparing the recipe with the following repo (all local deps get `0.0.1` version): https://github.com/novifinancial/winterfell

Emilgardis:
I've managed to make a minimal repro of this (or, a similar issue)

https://github.com/Emilgardis/cargo-chef-issue93

```
docker build .
Sending build context to Docker daemon  4.322MB
Step 1/17 : FROM lukemathwalker/cargo-chef:latest AS chef
 ---> cb6ae6b94ee0
Step 2/17 : RUN rustup default nightly
 ---> Using cache
 ---> 1894e37b5a40
[...]

PR #153, opened 2022-09-22
Add support for platform specific dependencies

sullivan-sean:
Cargo chef currently does not support local [platform specific dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies)

During the prepare stage, it does not update the version of a local dependency like
```toml
[target.'cfg(windows)'.dependencies]
local-windows-package = {path = "./path/to/local", version = "0.1.3" }
```
To
```toml
[target.'cfg(windows)'.dependencies]
local-windows-package = {path = "./path/to/local", version = "0.0.1" }
```

But all local dependencies are set to version `0.0.1`, which means there is a version mismatch and `cargo chef cook` fails with a message like in https://github.com/LukeMathWalker/cargo-chef/issues/93

This PR fixes this behavior by adhering directly to all of the possible ways to specify dependencies defined in `cargo`'s [documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

### DIFF
--- a/src/skeleton/version_masking.rs
+++ b/src/skeleton/version_masking.rs
@@ -57,25 +57,38 @@ fn mask_local_versions_in_manifests(
                 *version = toml::Value::String(CONST_VERSION.to_string());
             }
         }
-        mask_local_dependency_versions(local_package_names, manifest, "dependencies");
-        mask_local_dependency_versions(local_package_names, manifest, "dev-dependencies");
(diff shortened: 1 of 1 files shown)


LukeMathWalker (approved):
It looks good to me! Thanks for the contribution!

PR #154, opened 2022-09-26
feat: Add --timings argument

hobietje:
Pass through the `--timings` flag to `cargo` to be able to diagnose build times of dependencies.


LukeMathWalker (approved):
Thank you!

PR #157, opened 2022-10-12
Support multiple `--target` flags in one invocation

omninonsense:
Adds support for multiple `--target` flags on the command line (similar to how cargo now supports it)


LukeMathWalker (approved):
Thank you for the PR!

Issue #135, opened 2022-06-07
Compatibility with cargo-zigbuild


ghost:
Firstly thank you for this amazing project. I have used it for a long time and it's always worked well. It actually makes rust development in containers enjoyable. Recently I have had to build multiarch containers. I really like cargo-zigbuild

Is it possible to get this to work with cargo-chef?

LukeMathWalker:
Hey!
I have never used `cargo-zigbuild`. What would it entail to make it compatible?

crc-sstubbs:
I'm not completely sure what is happening during the cargo chef cook phase. Is this interacting with cargo build or is more complicated than that? If it is it would be a case of using cargo zigbuild instead of cargo build if a certain argument was added similar to how it currently works with musl. 

I have had quite a few issues with musl in the past and zigbuild seems to run better. Here is a dockerfile I have using cargo zigbuild. This is using docker cache mounts and works great locally but doesn't work nearly as well as cargo chef on CI. 

I'm not really sure how to help with this but would be happy to if I can.

```
# Builder
FROM --platform=$BUILDPLATFORM rust:1.62.0 AS builder
ARG BUILDPLATFORM
ARG ZIGVERSION=0.9.1
ARG ZIGBUILDVERSION=0.10.2
RUN apt-get update && apt-get install -y \
    ca-certificates \
    wget \
    && rm -rf /var/lib/apt/lists/*

# Zig
RUN if [ "${BUILDPLATFORM#*linux/}" = "amd64" ]; then \
    wget https://ziglang.org/download/${ZIGVERSION}/zig-linux-x86_64-${ZIGVERSION}.tar.xz && \
    tar -xf zig-linux-x86_64-${ZIGVERSION}.tar.xz && \
    rm -rf zig-linux-x86_64-${ZIGVERSION}.tar.xz && \
    ln -s /zig-linux-x86_64-${ZIGVERSION}/zig /usr/bin/; \
    else \
    wget https://ziglang.org/download/${ZIGVERSION}/zig-linux-aarch64-${ZIGVERSION}.tar.xz && \
    tar -xf zig-linux-aarch64-${ZIGVERSION}.tar.xz && \
    rm -rf zig-linux-aarch64-${ZIGVERSION}.tar.xz && \
    ln -s /zig-linux-aarch64-${ZIGVERSION}/zig /usr/bin/; \
    fi

# Cargo zigbuild
[...]

PR #158, opened 2022-10-12
Add support for `cargo-zigbuild`

omninonsense:
Adds support for [`cargo-zigbuild`](https://github.com/messense/cargo-zigbuild). This allows for better UX surrounding cross-platform compilation.

Users still need to install cargo-zigbuild manually and also install the Zig toolchain, though. I suppose it could be installed by chef too, but I'm not sure if that's the expected/desired behaviour.

Related:
 - https://github.com/LukeMathWalker/cargo-chef/issues/135

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -125,6 +125,10 @@ pub struct Cook {
     /// that are not necessary to successfully compile the specific binary.
     #[clap(long)]
     bin: Option<String>,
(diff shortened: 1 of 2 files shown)


LukeMathWalker (approved):
Thank you for the PR!
I admit I knew little about `cargo-zigbuild`, but it seems useful, so curious to give it a try 😁

PR #159, opened 2022-10-19
Fix passing --profile=(release/dev)

coolreader18:


coolreader18:
Oops, meant to provide a description: I had `ARG CARGO_PROFILE=release` in a Dockerfile and was passing it to `--profile=${CARGO_PROFILE}`, but when it's `dev` cargo-chef was assuming that it was to `target/dev` when the dev profile actually goes to `target/debug`.

LukeMathWalker (approved):
Thank you for the PR!

Issue #160, opened 2022-10-20
failed to select a version for the requirement `clap = "=4.0.15"`


erayerdin:
My Dockerfile is as below:

```
FROM lukemathwalker/cargo-chef:latest-rust-1.56.0 AS chef
WORKDIR /app

FROM chef AS planner
COPY . .
RUN cargo chef prepare --recipe-path recipe.json

FROM chef AS builder 
COPY --from=planner /app/recipe.json recipe.json
RUN cargo chef cook --release --recipe-path recipe.json
COPY . .
RUN cargo build --release --bin app

FROM debian:buster-slim AS runtime
WORKDIR /app
COPY --from=builder /app/target/release/app /usr/local/bin
ENTRYPOINT ["/usr/local/bin/app"]
```

But it fails saying:

```
    Updating crates.io index
error: failed to select a version for the requirement `clap = "=4.0.15"`
candidate versions found which didn't match: 3.2.22, 3.2.21, 3.2.20, ...
location searched: crates.io index
required by package `reactix v0.0.1 (/app)`
thread 'main' panicked at 'Exited with status code: 101', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.31/src/recipe.rs:145:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
The command '/bin/sh -c cargo chef cook --release --recipe-path recipe.json' returned a non-zero code: 101
```

LukeMathWalker:
Can you link the source code for the application?

erayerdin:
Seems like this is a problem about my program. So I'm closing this issue.

PR #161, opened 2022-10-20
Fix Docker Tag in Readme

erayerdin:
Readme refers to an old version of Rust. [Rust image](https://hub.docker.com/_/rust) already has `1` tag that refers to the latest stable version of 1.* for Rust or [cargo-chef image](https://hub.docker.com/r/lukemathwalker/cargo-chef/) already has `latest-rust-1` tag that refers to the same.

Some dependencies might require latest stable to work. In #160, `clap` 4.15.0 was a dependency, which was only available on the latest version of Rust. So, the build failed due to readme containing an old version of Rust.

Thus, it is sensible to change these lines in readme in order to refer to the latest stable version of Rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ cargo +nightly chef cook --recipe-path recipe.json
 `cargo-chef` is designed to be leveraged in Dockerfiles:
 
 ```dockerfile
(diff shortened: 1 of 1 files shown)


LukeMathWalker (approved):
Thanks!

PR #168, opened 2022-11-10
Add support for the .cargo/config configuration file.

liarokapisv:
This PR adds support for `.cargo/config` without an extension, as a valid configuration file. 

I implemented this after spending a couple of hours trying to understand why chef would not cache the dependencies during my builds. It turns out my project used the .cargo/config configuration file, leading into inconsistent builds between cargo and chef.


LukeMathWalker (approved):
TIL about `config`! Thank you 😁

PR #171, opened 2022-12-05
feat: mask workspace local dependencies

### LINKED ISSUE #167: Local workspace.dependencies versions are not being masked
The version masking of local dependencies is not being applied to local workspace dependencies that are defined in the root `Cargo.toml` under `[workspace.dependencies]`. This causes the version resolution to fail in `cargo cook`

https://github.com/LukeMathWalker/cargo-chef/blob/main/src/skeleton/version_masking.rs#L71
only handles dependencies under `target.*` and without any prefix. It should be extended by the `workspace` parent key.

oddgrd:
This PR applies version masking to local workspace dependencies. I ran into this when starting to implement workspace inheritance at shuttle: https://github.com/shuttle-hq/shuttle/pull/506. I tested it out by cloning my branch of cargo-chef in our Containerfile, and that caused the builds to succeed. 

This is my first PR to cargo-chef, so please let me know if I've misunderstood anything. :smile:  

Closes #167 

### DIFF
--- a/src/skeleton/version_masking.rs
+++ b/src/skeleton/version_masking.rs
@@ -83,7 +83,7 @@ fn mask_local_dependency_versions(
         }
     }
 
-    // There are two ways to specify dependencies:
+    // There are three ways to specify dependencies:
     // - top-level
     // ```toml
     // [dependencies]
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
That looks perfect, thanks for the contribution!

oddgrd:
My pleasure, thanks for the quick release!

PR #174, opened 2022-12-09
cache registry in builder stage with buildkit functionality

brandonros:
is this needed? does this help?


LukeMathWalker:
This is something you can do in your project, if you want to and happen to have something you can use to bootstrap the registry cache.
With the introduction of the sparse registry, which `cargo-chef` enables by default, the gains are likely to be minor.

PR #178, opened 2022-12-20
Support for custom target files

### LINKED ISSUE #177: Cargo chef cook fails on custom build targets
Given a custom target specified using a local JSON file, such as [this one](https://github.com/rust-wii/ogc-rs/blob/master/powerpc-unknown-eabi.json), when running e.g.

```
cargo +nightly chef cook --no-std --recipe-path recipe.json -Z build-std=core,alloc --target powerpc-unknown-eabi.json
```

Results in the following crash:

```
   Compiling some-dependency-name v0.0.1 (/app/lib)
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
Error: Failed to cook recipe.

Caused by:
    0: Failed to clean up dummy compilation artifacts.
    1: IO error for operation on /build/target/powerpc-unknown-eabi.json/debug: No such file or directory (os error 2)
    2: No such file or directory (os error 2)
```

It seems like in https://github.com/LukeMathWalker/cargo-chef/blob/37834d588b0f0dc792fc2f755060add906c4ffa9/src/skeleton/mod.rs#L241 the value of the `--target` argument is 1:1 used as folder name.

It is not strange that this feature is missing, as I believe it is a nightly feature. 
Nonetheless, its usage seems prevalent in embedded Rust: https://docs.rust-embedded.org/embedonomicon/custom-target.html .

As such, it would be nice if it could be supported. I'd happily contribute a PR for this.

Qqwy:
Fixes #177 

Changes:
- Support `--target=target-name.json` to be passed to `cargo chef`, by fixing the target location that is used during dummy cleanup in this case.

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -270,8 +270,6 @@ fn panic(_: &core::panic::PanicInfo) -> ! {
             .map(|path| path.join(&profile))
             .collect();
 
-        println!("target_directories: {:?}", target_directories);
-
         for manifest in &self.manifests {
             let parsed_manifest =
                 cargo_manifest::Manifest::from_slice(manifest.contents.as_bytes())?;
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -322,7 +322,6 @@ fn target_str(target: &String) -> &str {
     target.as_str().trim_end_matches(".json")
 }
 
-
 fn serialize_manifests(manifests: Vec<ParsedManifest>) -> Result<Vec<Manifest>, anyhow::Error> {
     let mut serialised_manifests = vec![];
     for manifest in manifests {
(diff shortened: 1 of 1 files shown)


Qqwy:
The PR does not yet contain a test for this functionality. I was not entirely sure where to start with writing such a test case if desired.

Since it is an unstable nightly feature, do we want a test or not?

LukeMathWalker (approved):

LukeMathWalker:
This is fine as is, happy to merge it.

Qqwy:
> This is fine as is, happy to merge it.

Great! Thank you!

I have pushed two new commits that should make rustfmt and clippy happy with the PR as well 😊

LukeMathWalker:
There is one more lint error, but I can take care of it. Thanks for the PR!

PR #190, opened 2023-02-09
ci: Fix failing builds (again)

### LINKED ISSUE #185: Docker build error:  could not find `cargo-chef` in registry `crates-io`
Hey Luca, first of all thank you very much for your hard work!

I'd like to use the docker image tagged as `latest-1.67-slim`, but it looks like slim image builds have been broken for a while. I believe this is the critical part of the build log:

```
#8 [linux/amd64 2/2] RUN ((cat /etc/os-release | grep ID | grep alpine) && apk add --no-cache musl-dev || true)     && cargo install cargo-chef --locked --version 0.1.51     && rm -rf /usr/local/cargo/registry/
#8 0.192     Updating crates.io index
#8 0.238 error: could not find `cargo-chef` in registry `crates-io` with version `=0.1.51`
#8 ...
```

Here's an example of a [failing job](https://github.com/LukeMathWalker/cargo-chef/actions/runs/4022526861/jobs/6912362787).

Do you think you could look into this?

jacobsvante:
Fixes #185

*Description of changes:*
Remove environment variable CARGO_NET_GIT_FETCH_WITH_CLI=true. [Gets rid of](https://github.com/jacobsvante/cargo-chef/actions/runs/4135633866/jobs/7148369885) the `could not find cargo-chef in registry crates-io with version =X.X.X` error message that all alpine and some debian docker image builds end up with.

I think this optimization will soon be completely unnecessary anyway, considering that [sparse-registry will be stabilized in the next rust release](https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html).

(Fun side note is that this is exactly the same PR I did [half a year ago](https://github.com/LukeMathWalker/cargo-chef/pull/144).)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,6 @@
 ARG BASE_IMAGE=rust
 FROM $BASE_IMAGE
 ARG CHEF_TAG
-ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 
 # Install musl-dev on Alpine to avoid error "ld: cannot find crti.o: No such file or directory"
 RUN ((cat /etc/os-release | grep ID | grep alpine) && apk add --no-cache musl-dev || true) \


jacobsvante:
It seems that there's some issue with the cargo-tarpaulin action (`Error: Couldn't find a release tarball containing binaries for latest`).

jacobsvante:
I noticed that the git tags weren't updated in my fork. So the workflow run that I linked to earlier is building an old version of cargo-chef. Now I'm doing a [new workflow run](https://github.com/jacobsvante/cargo-chef/actions/runs/4137423120/jobs/7152584521) for version 0.1.51, which hopefully will also yield positive results.

jacobsvante:
Okay all Docker builds are working now: https://github.com/jacobsvante/cargo-chef/actions/runs/4137423120/jobs/7152584521
Removing Draft status - this is ready for merge.

For others who are affected by this, you can use the latest builds here: https://hub.docker.com/r/jacobsvante/cargo-chef/tags?page=1&name=alpine

LukeMathWalker:
Thanks for having a look - I am quite short on time at the moment and I have been procrastinating on the investigation for a while.

PR #193, opened 2023-03-09
chore: Use sparse protocol to speed up builds

jacobsvante:
Should work for Rust 1.68 and upwards.

For more information, please see https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html.

Rust 1.68 is due to be released today. Once the Docker Hub tags for 1.68 exists this will be testable.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,7 @@
 ARG BASE_IMAGE=rust
 FROM $BASE_IMAGE
 ARG CHEF_TAG
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
Eager to test this out once the image is published!

LukeMathWalker:
Let's give it a go - builds are failing anyway due to how long it takes to update the git-based index.

jacobsvante:
Seems to be working pretty incredibly!

About 2.7 seconds from message `Updating crates.io index` until last `Downloaded proc-macro-error v1.0.4` (last `Downloaded` message)!!!

PR #202, opened 2023-04-11
Build linux-gnu binaries on ubuntu 20.04

### LINKED ISSUE #188: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /usr/local/bin/communication)
I used cargo chef to build my image, but my docker run fails with the message: 
```
/usr/local/bin/communication: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /usr/local/bin/communication)
```

Here is my Dockerfile: 
```
FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
WORKDIR communication

FROM chef AS planner
COPY . .
RUN cargo chef prepare --recipe-path recipe.json

FROM chef AS builder 
COPY --from=planner /communication/recipe.json recipe.json
# Build dependencies - this is the caching Docker layer!
RUN cargo chef cook --release --recipe-path recipe.json
# Build application
COPY . .
RUN cargo build --release --bin communication

# We do not need the Rust toolchain to run the binary!
FROM debian:buster-slim AS runtime
WORKDIR communication
COPY --from=builder /communication/target/release/communication /usr/local/bin
ENTRYPOINT ["/usr/local/bin/communication"]
```

I'm sorry I'm a bit of a noob to Rust. Wasn't sure if this problem had already been ran into?

Thanks!

Update: For extra context here's my Cargo.toml deps
```
async-trait = "0.1.58"
chrono = "0.4.23"
futures = "0.3.25"
gtmpl = "0.7"
openssl = { version = "0.10", features = ["vendored"] }
mongodb = { version = "2.3.1"}
phonenumber = "0.3.1"
regex = "1.3.9"
serde = "1.0.152"
serde_derive = "1.0.151"
serde_json = "1.0.89"
tokio = "1.22.0"
twilio-openapi = "1.2.2"
warp = "0.3"
tera = "1.17.1"
```

coolreader18:
*Issue #, if available:*
Fixes #188

*Description of changes:*
`ubuntu-latest` recently changed to be 22.04, from 20.04. This broke binaries that were being built being run in older distros (e.g., Debian stable, as is the default for the `rust` docker image). This fixes that by building the binaries against the older glibc.

### DIFF
--- a/.github/workflows/publish_binaries.yml
+++ b/.github/workflows/publish_binaries.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-latest
+            os: ubuntu-20.04
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
           - target: x86_64-apple-darwin


coolreader18:
Hm, that failure seems unrelated.

LukeMathWalker:
How long before Debian stable updates to the newer version of glibc?

coolreader18:
So it looks like the next debian stable ("bookworm") should be released sometime later this year; there's no set date yet but the release status page is [here](https://release.debian.org/). And not that a consistent pattern can necessarily be made out of this, but stretch was released in June 2017, buster was released in July 2019, and bullseye was released in August 2019. Also, bullseye and bookworm were interestingly both hard-frozen on the same date (March 23). So based off that I think we can expect at _least_ a few more months until it goes stable. Also though, debian oldstable is still fully supported for another 2 years, so just because the newest debian is out doesn't mean the previous one falls out of use.

LukeMathWalker:
Let's try to stay aligned with the `rust` image to minimise issues. We'll revise when they change the version of Debian they target with the `1` target.

Issue #152, opened 2022-09-06
Failure to load manifests for workspace members matched by glob


seanpianka:
I have a workspace that defines its members using a glob:
```toml
cargo-features = ["resolver"]

[workspace]
members = ["crates/*"]

# ...
```

When running `cook` after `prepare`, I see the following error output which complains that it cannot match any workspace member named `*`, rather than globbing for all members within the code-specific directory `crates/` of my project located at `/opt/api/`:
```
#19 [build 11/17] RUN cargo chef cook --recipe-path recipe.json
#19 sha256:c22e9955ddbcc98df6ecc1b985b2bfbbc0ab1d6e32028737a128bdca9569d249
#19 0.767 error: failed to load manifest for workspace member `/opt/api/crates/*`
#19 0.767
#19 0.767 Caused by:
#19 0.767   failed to read `/opt/api/crates/*/Cargo.toml`
#19 0.767
#19 0.767 Caused by:
#19 0.767   No such file or directory (os error 2)
#19 0.773 thread 'main' panicked at 'Exited with status code: 101', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.41/src/recipe.rs:161:27
#19 0.773 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
#19 ERROR: executor failed running [/bin/sh -c cargo chef cook --recipe-path recipe.json]: exit code: 101
```

Am I using this wrong, or is there still more work to be done on supporting workspaces?

LukeMathWalker:
That's correct, we do not supporting globbing syntax at the moment.

A PR would be welcome!

LukeMathWalker:
A fix has been released in 0.1.56 thanks to @Kobzol's work 🎉 
Can you confirm it works as expected @seanpianka?

nebnes:
I encounter the same error: 
```
 > [builder 2/4] RUN cargo chef cook --release --recipe-path recipe.json:
#0 0.498 error: failed to load manifest for workspace member `/app/./crate/*`
#0 0.498 
#0 0.498 Caused by:
#0 0.498   failed to read `/app/crate/*/Cargo.toml`
#0 0.498 
#0 0.498 Caused by:
#0 0.498   No such file or directory (os error 2)
[...]

PR #204, opened 2023-04-16
Use `cargo metadata` to find workspace manifests

Kobzol:
*Issue #, if available:*
Fixes: https://github.com/LukeMathWalker/cargo-chef/issues/152

*Description of changes:*
This PR adds basic usage of `cargo metadata`, which is not yet used to extract targets, but currently only to find workspace packages. This, adds built-in support for:
1) Ignoring vendored dependencies
2) Ignoring packages that are not part of the root workspace
3) Workspaces with packages specified using glob paths
4) Possibly also some other issues.

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -33,9 +33,11 @@ impl Skeleton {
         base_path: P,
         member: Option<String>,
     ) -> Result<Self, anyhow::Error> {
+        let metadata = extract_cargo_metadata(base_path.as_ref())?;
+
         // Read relevant files from the filesystem
         let config_file = read::config(&base_path)?;
-        let mut manifests = read::manifests(&base_path, config_file.as_deref())?;
+        let mut manifests = read::manifests(&base_path, metadata)?;
         if let Some(member) = member {
             ignore_all_members_except(&mut manifests, member);
         }
@@ -335,9 +337,16 @@ fn serialize_manifests(manifests: Vec<ParsedManifest>) -> Result<Vec<Manifest>, 
     Ok(serialised_manifests)
 }
 
+fn extract_cargo_metadata(path: &Path) -> Result<cargo_metadata::Metadata, anyhow::Error> {
+    let mut cmd = cargo_metadata::MetadataCommand::new();
+    cmd.current_dir(path);
+
+    cmd.exec().context("Cannot extract Cargo metadata")
+}
+
 /// If the top-level `Cargo.toml` has a `members` field, replace it with
 /// a list consisting of just the specified member.
-fn ignore_all_members_except(manifests: &mut [ParsedManifest], member: String) {
+fn ignore_all_members_except(manifests: &mut Vec<ParsedManifest>, member: String) {
     let workspace_toml = manifests
         .iter_mut()
         .find(|manifest| manifest.relative_path == std::path::PathBuf::from("Cargo.toml"));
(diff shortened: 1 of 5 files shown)


src/skeleton/mod.rs
```diff
@@ -346,6 +355,18 @@ fn ignore_all_members_except(manifests: &mut [ParsedManifest], member: String) {
         .and_then(|toml| toml.contents.get_mut("workspace"))
         .and_then(|workspace| workspace.get_mut("members"))
     {
-        *members = toml::Value::Array(vec![toml::Value::String(member)]);
+        *members = toml::Value::Array(vec![toml::Value::String(member.clone())]);
     }
+
+    // Remove manifest(s) with package name equal to `member`
```
LukeMathWalker, line 361: I think it's the opposite, isn't it?
We are *keeping* the manifests that don't have a package name or whose package name is *equal* to `member`.

src/skeleton/read.rs
```diff
@@ -35,87 +35,63 @@ pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, an
...
-        .build()
-        .context("Failed to scan the files in the current directory.")?;
+    let mut metadata_paths = metadata
+        .workspace_packages()
+        .into_iter()
+        .map(|package| package.manifest_path.clone().into_std_path_buf())
+        .collect::<HashSet<_>>();
+    metadata_paths.insert(base_path.as_ref().join("Cargo.toml"));
+    let mut manifest_paths: Vec<PathBuf> = metadata_paths.into_iter().collect();
+    manifest_paths.sort();
```
LukeMathWalker, line 49: We can avoid this "collect" + "collect" + "sort" by using a `BTreeSet` directly.

tests/skeletons.rs
```diff
@@ -530,7 +527,7 @@ name = "test-dummy"
 path = "src/main.rs"
 
 [dependencies]
-uuid = { version = "=0.8.0", features = ["v4"] }        
+either = { version = "=1.8.1" }
```
LukeMathWalker, line 530: Same here—why the swap?

LukeMathWalker:
Looking really good!
Left a few comments, but nothing major.

src/skeleton/mod.rs
Kobzol: Of course, the comment is opposite, I'll fix it.

tests/skeletons.rs
```diff
@@ -521,7 +518,7 @@ members = [
             "src/project_a",
             r#"
 [package]
-name = "project-a"
+name = "project_a"
```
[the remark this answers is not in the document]
Kobzol: I changed it for consistency, the second project was named `project_b` (my OCD was triggered I suppose :sweat_smile:). I can revert it if you want.

src/skeleton/read.rs
```diff
@@ -35,87 +35,63 @@ pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, an
...
-fn vendored_directory(config_contents: Option<&str>) -> Option<String> {
-    let contents = config_contents.and_then(|contents| contents.parse::<toml::Value>().ok())?;
-    let source = contents.get("source")?;
-    let crates_io = source.get("crates-io")?;
-    let vendored_field_suffix = crates_io
-        .get("replace-with")
-        .and_then(|value| value.as_str())?;
-    let vendored_sources = source.get(vendored_field_suffix)?;
-    Some(vendored_sources.get("directory")?.as_str()?.to_owned())
-}
```
LukeMathWalker, line 47: Very very happy of getting rid of this, it always felt like a bit of a hack! ❤️

tests/skeletons.rs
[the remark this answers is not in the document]
LukeMathWalker: IIRC it was intentional, to make sure we can handle the "non-normalized" version of a crate name.

tests/skeletons.rs
Kobzol: `uuid` had multiple dependencies, which weren't reflected in the lockfile of the original test, and it thus couldn't be parsed with `cargo metadata`. I tried to find a package without any dependencies, to keep the lockfile content simple, otherwise it would be quite large, which in turn made the test workspace setup unnecessarily large.

tests/skeletons.rs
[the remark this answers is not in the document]
Kobzol: Ok, I'll revert it and add a comment.

Kobzol:
Pushed a commit that should resolve the review comments. I'll rebase it before merging if you want.

LukeMathWalker:
No need, I squash and merge anyway. Good to go!

PR #207, opened 2023-04-19
We should keep all manifests, even if they are not mentioned in the list of workspace members

LukeMathWalker:
…since the binary target might depend on them via a path dependency.

We could be more surgical and keep exclusively the ones that are in the binary's transitive dependency closure, but we'll leave that for a future PR.

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -357,16 +357,4 @@ fn ignore_all_members_except(manifests: &mut Vec<ParsedManifest>, member: String
     {
         *members = toml::Value::Array(vec![toml::Value::String(member.clone())]);
     }
-
-    // Keep only manifest(s) with package name equal to `member`
-    manifests.retain(|manifest| {
-        let package_name = manifest
-            .contents
-            .as_table()
-            .and_then(|t| t.get("package"))
-            .and_then(|t| t.as_table())
-            .and_then(|t| t.get("name"))
-            .and_then(|t| t.as_str());
-        package_name.is_none() || package_name == Some(&member)
-    });
 }
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -346,7 +346,7 @@ fn extract_cargo_metadata(path: &Path) -> Result<cargo_metadata::Metadata, anyho
 
 /// If the top-level `Cargo.toml` has a `members` field, replace it with
 /// a list consisting of just the specified member.
-fn ignore_all_members_except(manifests: &mut Vec<ParsedManifest>, member: String) {
+fn ignore_all_members_except(manifests: &mut [ParsedManifest], member: String) {
     let workspace_toml = manifests
         .iter_mut()
         .find(|manifest| manifest.relative_path == std::path::PathBuf::from("Cargo.toml"));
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
I'm aware—the expected behaviour is removing it from the list of workspace members but _keeping_ the actual manifest of that crate around, so that it's picked up as a path dependency. 

Can you test again on `0.1.57`? @Zizico2

oddgrd:
Hey @LukeMathWalker! We are also experiencing this issue, and it persists on `0.1.57`. Like @Zizico2 it also works for us if we use `0.1.54`. 
Dockerfile: https://github.com/shuttle-hq/shuttle/blob/main/Containerfile
Workspace manifest: https://github.com/shuttle-hq/shuttle/blob/main/Cargo.toml

Error:
```
 > [builder 2/4] RUN cargo chef cook $(if [ "debug" = "release" ]; then echo --debug; fi) --recipe-path recipe.json:
#0 0.456 error: failed to get `shuttle-secrets` as a dependency of package `cargo-shuttle v0.0.1 (/build/cargo-shuttle)`
#0 0.456
#0 0.456 Caused by:
#0 0.456   failed to load source for dependency `shuttle-secrets`
#0 0.456
#0 0.456 Caused by:
#0 0.456   Unable to update /build/resources/secrets
#0 0.456
#0 0.456 Caused by:
#0 0.456   failed to read `/build/resources/secrets/Cargo.toml`
#0 0.456
#0 0.456 Caused by:
#0 0.456   No such file or directory (os error 2)
#0 0.459 thread 'main' panicked at 'Exited with status code: 101', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.57/src/recipe.rs:189:27
#0 0.459 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
------
Containerfile:32
--------------------
  30 |     COPY --from=planner /build/recipe.json recipe.json
  31 |     ARG CARGO_PROFILE
  32 | >>> RUN cargo chef cook $(if [ "$CARGO_PROFILE" = "release" ]; then echo --${CARGO_PROFILE}; fi) --recipe-path recipe.json
  33 |     COPY --from=cache /build .
  34 |     ARG folder
--------------------
ERROR: failed to solve: process "/bin/sh -c cargo chef cook $(if [ \"$CARGO_PROFILE\" = \"release\" ]; then echo --${CARGO_PROFILE}; fi) --recipe-path recipe.json" did not complete successfully: exit code: 101
make: *** [Makefile:162: shuttle-provisioner] Error 1
```

Zizico2:
> I'm aware—the expected behaviour is removing it from the list of workspace members but _keeping_ the actual manifest of that crate around, so that it's picked up as a path dependency.
> 
> Can you test again on `0.1.57`? @Zizico2

It worked @LukeMathWalker

LukeMathWalker:
Your setup is quite a bit more intricate @oddgrd, but I think the root cause is the fact that you are excluding that crate (`shuttle-secrets`) from the workspace [here](https://github.com/shuttle-hq/shuttle/blob/4ce62aa5eeff1d4673c7d60b57734a636ee30c46/Cargo.toml#L21).

It should be enough to either:
- Stop excluding it;
- `COPY` the `resources` folder before running `cargo chef cook`.

PR #210, opened 2023-04-21
fix: Run `cargo metadata` in offline mode

asandroq:
This restores the old behaviour of ``cargo chef prepare`` which would finish quickly and not need Git credentials for private repos.


LukeMathWalker:
I'm afraid this would cause the command to fail unless crates had been downloaded beforehand.

asandroq:
Yes, I've just realised that :-(

asandroq:
I've creates a new one, which seems to work: https://github.com/LukeMathWalker/cargo-chef/pull/211

PR #211, opened 2023-04-21
fix: Run `cargo metadata` with no-deps flag

asandroq:
This change restores the old behaviour of ``cargo chef prepare`` which would finish quickly and not need Git credentials for private repos.


Kobzol:
@LukeMathWalker I think that this should be fine.

PR #214, opened 2023-05-04
Use `cargo metadata` to resolve targets

Kobzol:
This should help generate the correct entrypoint file paths, instead of guessing them.

### DIFF
--- a/src/skeleton/read.rs
+++ b/src/skeleton/read.rs
@@ -2,8 +2,9 @@
 use super::ParsedManifest;
 use crate::skeleton::target::{Target, TargetKind};
 use cargo_metadata::{Metadata, Package};
+use std::collections::BTreeSet;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 pub(super) fn config<P: AsRef<Path>>(base_path: &P) -> Result<Option<String>, anyhow::Error> {
@@ -39,27 +40,29 @@ pub(super) fn manifests<P: AsRef<Path>>(
     base_path: &P,
     metadata: Metadata,
 ) -> Result<Vec<ParsedManifest>, anyhow::Error> {
-    let mut packages = metadata
+    let mut packages: BTreeSet<(BTreeSet<Target>, PathBuf)> = metadata
         .workspace_packages()
         .iter()
         .copied()
         .chain(metadata.root_package())
-        .map(|p| (Some(p), p.manifest_path.clone().into_std_path_buf()))
-        .collect::<Vec<_>>();
+        .map(|p| {
+            (
+                gather_targets(p),
+                p.manifest_path.clone().into_std_path_buf(),
+            )
+        })
+        .collect::<BTreeSet<_>>();
 
     if metadata.root_package().is_none() {
         // At the root, there might be a Cargo.toml manifest with a [workspace] section.
         // However, if this root manifest doesn't contain [package], it is not considered a package
         // by cargo metadata. Therefore, we have to add it manually.
         // Workspaces currently cannot be nested, so this should only happen at the root.
-        packages.push((None, base_path.as_ref().join("Cargo.toml")));
+        packages.insert((Default::default(), base_path.as_ref().join("Cargo.toml")));
     }
 
-    packages.sort_by(|a, b| a.1.cmp(&b.1));
-    packages.dedup();
-
     let mut manifests = vec![];
-    for (package, absolute_path) in packages {
+    for (targets, absolute_path) in packages {
         let contents = fs::read_to_string(&absolute_path)?;
 
         let mut parsed = cargo_manifest::Manifest::from_str(&contents)?;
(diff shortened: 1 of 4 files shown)


src/skeleton/read.rs
```diff
@@ -39,20 +39,27 @@ pub(super) fn manifests<P: AsRef<Path>>(
...
+    if metadata.root_package().is_none() {
+        // At the root, there might be a Cargo.toml manifest with a [workspace] section.
+        // However, if this root manifest doesn't contain [package], it is not considered a package
+        // by cargo metadata. Therefore, we have to add it manually.
+        // Workspaces currently cannot be nested, so this should only happen at the root.
+        packages.push((None, base_path.as_ref().join("Cargo.toml")));
+    }
+
+    packages.sort_by(|a, b| a.1.cmp(&b.1));
+    packages.dedup();
```
LukeMathWalker, line 59: Can't we use a `BTreeSet` directly here?

LukeMathWalker: We just to introduce a small struct and derive the relevant traits, which will give us named fields along the way, making the code a bit clearer as well.

src/skeleton/read.rs
```diff
@@ -89,14 +96,62 @@ pub(super) fn manifests<P: AsRef<Path>>(
                 &absolute_path
             )
         })?;
+        let mut targets = package.map(|p| gather_targets(p)).unwrap_or_default();
+        targets.sort_by(|a, b| a.path.cmp(&b.path));
```
LukeMathWalker, line 100: We can use a `BTreeSet` here as well.

src/skeleton/read.rs
```diff
@@ -89,14 +96,62 @@ pub(super) fn manifests<P: AsRef<Path>>(
...
             contents: intermediate,
+            targets,
         });
     }
 
     Ok(manifests)
 }
+// I started implementing resolving targets through `cargo metadata`, but I wonder if it makes sense. We currently recreate
+fn gather_targets(package: &Package) -> Vec<Target> {
+    let manifest = package.manifest_path.clone().into_std_path_buf();
```
LukeMathWalker, line 113:
```suggestion
    let manifest_path = package.manifest_path.clone().into_std_path_buf();
```

src/skeleton/read.rs
```diff
@@ -89,14 +96,62 @@ pub(super) fn manifests<P: AsRef<Path>>(
...
+    let root_dir = manifest.parent().unwrap();
+    package
+        .targets
+        .iter()
+        .filter_map(|target| {
```
LukeMathWalker, line 118: Isn't this just a `map`?

src/skeleton/read.rs
Kobzol: With the tuple, probably not, since Package is not orderable. With a custom struct, sure :)

Kobzol:
```rust
struct PackageMetadata {
    package: Option<Package>,
    manifest_path: PathBuf,
}

impl PartialEq<Self> for PackageMetadata {
    fn eq(&self, other: &Self) -> bool {
        self.manifest_path.eq(&other.manifest_path)
    }
}

impl Eq for PackageMetadata {}

impl PartialOrd for PackageMetadata {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        self.manifest_path.partial_cmp(&other.manifest_path)
    }
}

impl Ord for PackageMetadata {
    fn cmp(&self, other: &Self) -> Ordering {
        self.manifest_path.cmp(&other.manifest_path)
    }
}
```
I'm not sure how to make this shorter. Is it worth it?

src/skeleton/read.rs
```diff
@@ -89,14 +96,62 @@ pub(super) fn manifests<P: AsRef<Path>>(
...
         manifests.push(ParsedManifest {
             relative_path,
             contents: intermediate,
+            targets,
         });
     }
 
     Ok(manifests)
 }
+// I started implementing resolving targets through `cargo metadata`, but I wonder if it makes sense. We currently recreate
```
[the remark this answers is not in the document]
Kobzol: Oops, that's a forgotten Zulip message :sweat_smile:

src/skeleton/read.rs
Kobzol: By implementing `Ord` for `Target`? I'm not sure if we want it to be orderable by path everywhere. Otherwise we'd need to wrap it in another newtype.

src/skeleton/read.rs
Kobzol: It is, good catch. It was returning `None` before, but I changed it.

src/skeleton/read.rs
LukeMathWalker: We don't actually care about _how_ they are ordered, we just need them to be ordered in order to have a deterministic output.

So you can derive all traits without bothering with a manual implementation.
IIRC it's enough to put the path as the first field to have them sorted by path and then by the contents of the package.

src/skeleton/read.rs
LukeMathWalker: What's the downside? They are all private implementation details of `cargo-chef`, nothing here is offered as a library to anyone else.

LukeMathWalker: And, same thing, it doesn't need to be sorted by path. It just needs to be sorted for determinism.

src/skeleton/read.rs
Kobzol: I can't use derive, since Package isn't Ord. Good point with the order not being really important, although for tests it's nice if it's by path I suppose.

LukeMathWalker: I see—you could get around it by grabbing the targets here, instead of doing it below.

Kobzol:
I added a few `Ord` derives and used `BTreeSets` to avoid manual sorting.

src/skeleton/read.rs
```diff
@@ -39,27 +40,29 @@ pub(super) fn manifests<P: AsRef<Path>>(
...
         .chain(metadata.root_package())
-        .map(|p| (Some(p), p.manifest_path.clone().into_std_path_buf()))
-        .collect::<Vec<_>>();
+        .map(|p| {
+            (
+                gather_targets(p),
+                p.manifest_path.clone().into_std_path_buf(),
+            )
+        })
+        .collect::<BTreeSet<_>>();
```
LukeMathWalker: Re-reading it now, I think we can go one step further and make this a `BTreeMap<PathBuf, BTreeSet<Target>>`, which better conveys the actual intent/constraints here.

Kobzol: Done.

LukeMathWalker (approved):

PR #221, opened 2023-06-08
Solve CI warnings replacing outdated actions

andros21:
**Issue #, if available**

Try to remove noisy warnings inside github actions

**Description of changes**

Principally by bumping to new tag version (e.g.
`action/checkout`). But for unmaintained actions the swap
with a fork or something similar, is the only solution:
* [`action-rs/toolchain`](https://github.com/actions-rs/toolchain/issues/216) -> [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain)
* [`action-rs/audit-check`](https://github.com/actions-rs/audit-check/issues/227) -> [`rustsec/audit-check`](https://github.com/rustsec/audit-check)

Two consequences of using `dtolnay/rust-toolchain` are:
* no more needed `CARGO_TERM_COLOR=always` already set by
the action
* use directly cargo commands in run steps

I also change [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) `::set-output` commands

A question to complete the PR:
How about `actions-rs/tarpauli` step? It was responsible for
the failure of general "Rust" workflow before the fork. It's
also unmaintained as all `action-rs/*`. A possible
replacement can be found, but I don't understand if it's still
used or not

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/.github/workflows/audit-on-push.yml
+++ b/.github/workflows/audit-on-push.yml
@@ -8,7 +8,7 @@ jobs:
   security_audit:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions-rs/audit-check@v1
+      - uses: actions/checkout@v3
+      - uses: rustsec/audit-check@v1.4.1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,7 +21,7 @@ jobs:
       -
         id: set-matrix
         run: |
-          echo -n '::set-output name=matrix::[' \
+          (echo -n 'matrix=[' \
           && curl --silent https://raw.githubusercontent.com/docker-library/official-images/master/library/rust \
             | grep -E Tags: \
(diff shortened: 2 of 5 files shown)


andros21:
I've pushed a proposal for fix it, using [`taiki-e/install-action`](https://github.com/taiki-e/install-action) to install `tarpaulin` from github-releases of the author, avoiding crate build

Let me know if there is something more I can do here.
Partly related:
1. to keep actions updated, how about enabling something like [dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)?
2. can be advisable to set explicitly `permissions` related to `GITHUB_TOKEN`,
for example in "Publish binaries" `permissions: {contents: write}`

If appreciated I can make these changes in a separate PR

LukeMathWalker:
I'd definitely be open to both changes.
On `dependabot`: I'd prefer a configuration that doesn't raise PRs unless there is a semver-breaking version bump.

PR #223, opened 2023-06-16
Best practies for github actions: permissions and dependabot

andros21:
**Related:** #221

**Description of changes:**

Explicitly set permissions required inside github actions

For **dependabot** what you'd prefer for `schedule` interval?
And for the preference you expressed:
> raise PRs unless there is a semver-breaking version bump

I think the closest behavior can be ignoring semver patch release for all actions,
instead minor, major usually can contain some breaking changes. Is it ok?

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-patch"
--- a/.github/workflows/audit-on-push.yml
+++ b/.github/workflows/audit-on-push.yml
@@ -4,6 +4,11 @@ on:
     paths:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
+
+permissions:
+  issues: write
+  checks: write
+
 jobs:
   security_audit:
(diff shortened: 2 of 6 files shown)


LukeMathWalker:
I think we can lean on "increase-if-necessary" for Dependabot? (see [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy))

LukeMathWalker:
Yeah, you're right. Let's stick to your suggestion then.

andros21:
To benefit from dependabot update checks the last thing that remain to do is to explicitly pin the complete tag version for all github actions used (to avoid rolling tags such as `checkout@v3`) even better the full commit hash.

Fortunately for this error prone task we can use https://github.com/step-security/secure-repo, I've already preview changes for full length commit sha pinning: https://github.com/step-security-bot/cargo-chef/commit/760092bca981d81aa9d3cb35a7678b17b4c3a55b

If you are ok with this approach, I can let the bot open a PR, and then incorporate its changes here

LukeMathWalker (approved):

LukeMathWalker:
I prefer to keep a rolling tag, which is why I want to skip patch updates (and perhaps minor, depending on the frequency).
Pinning would just lead to unnecessary PR churn on this repository.

Issue #182, opened 2023-01-10
Feature request: consider building docker images for armv7 architecture


beeb:
Thanks a lot for this great tool!

I was wondering if you would consider building multiarch images for `linux/arm/v7`additionally to the existing `linux/arm64` and `linux/amd64`? Since you are using buildx in CI with QEMU, I image it should work.

Thanks in advance.

dumbasPL:
Just a note to anyone else interested in armv7. Check out the [Without the pre-built image](https://github.com/LukeMathWalker/cargo-chef#without-the-pre-built-image) section of the readme. But I still think having prebuilt images would be nice.

Also, I've confirmed that simply adding `linux/arm/v7` to the CI works just fine (except of course for the Alpine images that we would need to exclude)

LukeMathWalker:
Happy to merge a PR in this direction 👌🏻

MyWay:
Any news on this?

PR #226, opened 2023-07-10
Support Docker image for i386 and arm32v7 architectures

MPThLee:
*Issue #, if available:* #182 

*Description of changes:* Add 386 and arm/v7 to Docker image build platform.

Official rust docker image is depends on rustup. And rustup doesn't support i386 and arm/v7 for musl(alpine) target.
This PR adds build target 'i386' and 'arm/v7' on non-alpine target. According to https://github.com/LukeMathWalker/cargo-chef/issues/182#issuecomment-1616805177, this should be ok.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -94,13 +94,14 @@ jobs:
           CHEF_PACKAGE_VERSION=${{ steps.package_version.outputs.result }}
           CHEF_IMAGE=$DOCKER_REPO:$CHEF_PACKAGE_VERSION-rust-$RUST_IMAGE_TAG
           CHEF_IMAGE_LATEST=$DOCKER_REPO:latest-rust-$RUST_IMAGE_TAG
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -119,6 +120,6 @@ jobs:
             --tag $DOCKER_REPO:latest \
             --build-arg=BASE_IMAGE=rust:$RUST_IMAGE_TAG \
             --build-arg=CHEF_TAG=$CHEF_PACKAGE_VERSION \
-            --platform linux/amd64,linux/arm64 \
+            --platform linux/amd64,linux/arm64,linux/386,linux/arm/v7 \
```
LukeMathWalker, line 123: Don't we have the same problem here with respect to Alpine?

MPThLee: latest should always be Debian, isn’t it?

Issue #238, opened 2023-08-28
Add `--bins` flag


mladedav:
We have a workspace with several projects, each containing one binary.

We would like to have a single caching layer where all dependencies for all binaries are already precompiled and then compile only the binaries themselves so that there aren't multiple cache docker layers.

Another thing I have noticed is `cargo chef` accepts only one `--bin` argument while `cargo build` et al. allow for multiple binaries to be specified. Either of those would help in our situation if you have reservations about either.

Originally I wanted to do just simple `cargo chef cook` and `cargo build -p package`, but it turns out this leads to several recompilations, seemingly because of feature sets and possibly other issues. `cargo chef cook` followed by `cargo build --bin package` is somewhat better, but there were still a few recompilations (although I didn't check closer why yet). The only way currently available I have found is `cargo chef cook --all-targets` but that of course also compiles a lot of stuff we do not want.

thomaseizinger:
I am running into the same issue and tried to use `--all-targets` but that didn't really help at all. I've contemplated to just invoke `cargo chef cook` multiple times, one for each binary. Until #241 is being released, would that be equivalent?

PR #241, opened 2023-09-06
Add `--bins` and allow `--bin` to be called multiple times

mladedav:
*Issue #, if available:* #238 

*Description of changes:* Adds `--bins` which builds all binaries. The second commit makes it possible to call `--bin` multiple times like `--package`. This is to be closer in usage to other cargo commands such as cargo build.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -138,10 +138,9 @@ pub struct Cook {
     /// Cook using `#[no_std]` configuration  (does not affect `proc-macro` crates)
     #[clap(long)]
     no_std: bool,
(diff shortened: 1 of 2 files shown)


mladedav:
@LukeMathWalker Sorry, I've just found this is still hanging here, can you please review this?

LukeMathWalker:
I'll try to get to it in the coming weeks, but my backlog is quite full at the moment.

LukeMathWalker:
Nothing to comment on, thank you for the PR!

PR #249, opened 2023-11-15
Update deps and remove sys dependencies

Eraden:
*Description of changes:*

Remove OS glibc dependency so it can be run on any linux without MUSL

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,64 +4,72 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.1.2"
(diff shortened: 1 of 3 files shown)


LukeMathWalker:
> It currently works on Nightly Rust on Linux on x86-64, x86, aarch64, and riscv64.

> It's not as complete as the major libc implementations. It can run most Rust code, and some popular C libraries, but still lacks a lot of things used by typical C code.

> It currently depends on Rust Nightly and only runs on Linux, and currently only on x86-64, x86, aarch64, and riscv64.

Not worth it yet, I'm sorry.

PR #255, opened 2024-02-14
use the manifest path instead of just the member name when using `prepare --bin`

ClementTsang:
*Issue #, if available:*

Closes: #232 

*Description of changes:*

The current code assumes that when running `prepare` with `--bin`, that we can remove all members and just add one member that is equal to whatever we passed in `--bin`.

The problem with that is that if your packages aren't in the exact same root directory, then this will result in a recipe file that breaks once you try calling `cook` on it.

To fix this, this PR instead sets the sole member to the _path_ of the package, which (I think) can be found using in `cargo_metadata::Metadata`, grabbing the manifest path, and setting it to be relative to the workspace root + removing `Cargo.toml` from the path. I chose to do it this way rather than just matching + grabbing whatever's in the `members` array as I don't think that would work due to wildcards being valid.

This also adds a test emulating various package locations in a workspace and seeing if the path is correctly set. This previously failed, but now passes with these changes. All other previously-existing tests seem to pass, so I don't believe this should break things. That said, if people were using the workaround mentioned in #232, then I believe this will be a breaking change, so that'll probably have to be communicated.

---

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -6,6 +6,7 @@ use crate::skeleton::target::{Target, TargetKind};
 use crate::OptimisationProfile;
 use anyhow::Context;
 use cargo_manifest::Product;
+use cargo_metadata::Metadata;
 use fs_err as fs;
 use globwalk::GlobWalkerBuilder;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


src/skeleton/mod.rs
```diff
@@ -294,18 +295,41 @@ fn extract_cargo_metadata(path: &Path) -> Result<cargo_metadata::Metadata, anyho
...
+                .into_iter()
+                .find(|pkg| pkg.name == member)
+            {
+                // Make this a relative path to the workspace, and remove the `Cargo.toml` child.
+                // XXX: Do we need to strip the prefix? Or can we use an absolute path?
+                let new_member_path = pkg
+                    .manifest_path
+                    .strip_prefix(workspace_root)
+                    .ok()
+                    .and_then(|mp| mp.parent());
```
LukeMathWalker, line 326: Let's use https://docs.rs/pathdiff/latest/pathdiff/fn.diff_paths.html here. `pathdiff` is already in the dependency tree.

LukeMathWalker:
Thank you for the excellent PR ❤️ 
I only have one comment related to path manipulation. Once that is solved we can merge.

ClementTsang:
Rebased to the newest `main` and cleaned up a bit along with the suggestion.

PR #247, opened 2023-09-28
Do not mask workspace dependencies with a source attribute

elliotkennedy:
Workspace dependencies with a `source` attribute are likely to be transitive and not sourced from the workspace.

*Issue #, if available:*

None raised.

*Description of changes:*

Similar to https://github.com/LukeMathWalker/cargo-chef/issues/224.

Given a workspace project:

```md
.
├── project_a
│   └── Cargo.toml
├── project_b
│   └── Cargo.toml
├── Cargo.lock
├── Cargo.toml
```

`project_a/Cargo.toml`:
```toml
[package]
name = "project_a"
version = "2.2.2"

[dependencies]
either = { version = "=1.8.1" }
```

`project_b/Cargo.toml`:
```toml
[package]
name = "project_b"
version = "3.3.3"

[dependencies]
# none
```

The external project, `either`, depends on a crates.io-published version of `project_b`. 

Therefore `project_b` is a transitive dependency of `project_a`, however the transitive `project_b` is different from the workspace version.

Currently `crates.io:project_b` is incorrectly masked to version 0.0.1 in `Cargo.lock`.

This PR identifies transitive dependencies by not masking dependencies with a `source` attribute. I think this is sensible and cannot think of a scenario where a workspace package would have a `source` attribute.

```toml
[[package]]
name = "either"
version = "1.8.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "7fcaabb2fef8c910e7f4c7ce9f67a1283a1715879a7c230ca9d6d1ae31f16d91"
dependencies = [
 "project_b 3.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
]

[[package]]
name = "project_a"
version = "0.0.1"
dependencies = [
 "either",
]

[[package]]
name = "project_b"
version = "0.0.1"

[[package]]
name = "project_b"
version = "3.3.3"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "7fcaabb2fef8c910e7f4c7ce9f67a1283a1715879a7c230ca9d6d1ae31f16d91"
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #263, opened 2024-03-18
Do not mask dependencies without `path` set

### LINKED ISSUE #224: package <packageName> is specified twice in the lockfile
This is possibly the same issue as #179 

We have a project with a workspace with crate `X`. One of the projects depends on an older version of `X` for compatibility (so it uses local `X` as well as an older version from crates).

When I run `cargo chef cook`, the lock file gets rewritten and both packages get their versions changed to `0.0.1` so cargo then complains that the same package is there twice.

I think the correct thing to do here would be to check not only the name but also the version of a package before changing it in Cargo.lock. I think that should be sufficient since there can not be the same package with multiple versions (e.g. one local one from crates).

mladedav:
*Issue #, if available:*

Fixes #224

Similar to #247 but for `Cargo.toml` manifests instead of `Cargo.lock`

*Description of changes:*

Mask only dependencies that are local, i.e. have `path` set. If they do not have this field set, they should be pulled from some registry.

This is not completely bulletproof as a local dependency could theoretically also point to a package outside a workspace which would have the same name as a package in the workspace, but this scenario is somewhat unlikely.


LukeMathWalker:
Thanks!

Issue #231, opened 2023-07-25
Put info about Rust toolchain into recipe and use it in `cargo chef cook`


haaawk:
# Problem
1. Currently `cargo chef cook` uses default Rust toolchain.
2. The project can define a Rust toolchain it wants to be built with using `rust-toolchain.toml`.
3. This may lead to dependencies being built using different toolchain than the actual project. As a result, `cargo build` will compile the dependencies again using the toolchain defined in `rust-toolchain.toml`.

The solution would be to include info about the toolchain into the recipe and use that info in cargo chef cook to make sure it compiles dependencies using the correct toolchain.

# Problem reproducer:
1. `cargo new --lib dummy`
2. cd dummy
3. Create `rust-toolchain.toml` with following content:
``` 
[toolchain]
profile = "default"
channel = "1.70.0"
```
4. Add `tokio` as a dependency in `Cargo.toml` to be able to observe it being compiled twice:
```
[package]
name = "dummy"
version = "0.1.0"
edition = "2021"

[dependencies]
tokio = { version = "1", features = ["full"] }
```
5. Create a `Dockerfile`:
```
FROM rust:slim-bullseye AS chef
RUN cargo install cargo-chef

FROM chef AS planner
COPY . .
RUN cargo chef prepare --recipe-path recipe.json

FROM chef AS builder
COPY --from=planner /recipe.json recipe.json
RUN cargo chef cook --recipe-path recipe.json
COPY . .
RUN cargo build
```
6. Run `docker build --progress=plain --target builder --no-cache . 2>&1 | tee build-wrong.log`
7. Observe dependencies being compiled both by `cargo chef cook` and `cargo build`:
```

...

#10 [builder 2/4] RUN cargo chef cook --recipe-path recipe.json
#10 0.334     Updating crates.io index
#10 0.699  Downloading crates ...
#10 1.053   Downloaded parking_lot_core v0.9.8
#10 1.058   Downloaded bytes v1.4.0
#10 1.064   Downloaded signal-hook-registry v1.4.1
#10 1.065   Downloaded num_cpus v1.16.0
#10 1.072   Downloaded lock_api v0.4.10
#10 1.076   Downloaded tokio-macros v2.1.0
#10 1.080   Downloaded scopeguard v1.2.0
#10 1.083   Downloaded pin-project-lite v0.2.10
#10 1.095   Downloaded autocfg v1.1.0
#10 1.097   Downloaded socket2 v0.4.9
[...]

PR #281, opened 2024-10-10
fix: use toolchain file while building dependencies

### LINKED ISSUE #271: toolchain in recipe.json is ignored in `cargo chef cook`
`cargo chef cook` generates rust-toolchain.toml file, but it's not used in the subsequent build of dependencies.
rust-toolchain.toml takes effect when  `cargo` is invoked next time. Therefore, executing `cargo chef cook` twice results in building with the wrong toolchain for the first time and the correct building for the second time.

The following Dockerfile reproduces the problem.

```Dockerfile
FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
WORKDIR /app

FROM chef AS planner
RUN echo '[workspace]\nresolver = "2"\nmembers = ["hello"]' > Cargo.toml
RUN cargo new hello
RUN echo '[toolchain]\nchannel = "1.75.0"' > rust-toolchain.toml
RUN cargo chef prepare --recipe-path recipe.json

FROM chef AS builder
COPY --from=planner /app/recipe.json recipe.json
RUN cargo chef cook --release --recipe-path recipe.json # does not install 1.75.0
RUN cargo chef cook --release --recipe-path recipe.json # does install 1.75.0
```

LtdSauce:
This commit make the cargo build command that builds the dependencies actually adhere to the toolchain file if it was present in the recipe.

Before this it was ignored and thus the next call to cargo build would rebuild all dependencies again, but then with what is actually contained in the toolchain file. This was because the environment variable 'RUSTUP_TOOLCHAIN' was already set and so the toolchain file was not considered, see here for details:
https://rust-lang.github.io/rustup/overrides.html

This should fix #271 (and also solve the missing piece for #231)

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -1,4 +1,4 @@
-use crate::Skeleton;
+use crate::{RustToolchainFile, Skeleton};
 use anyhow::Context;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
@@ -62,7 +62,7 @@ impl Recipe {
         if args.no_build {
             return Ok(());
         }
-        build_dependencies(&args);
+        build_dependencies(&args, &self.skeleton.rust_toolchain_file);
         self.skeleton
             .remove_compiled_dummies(
                 current_directory,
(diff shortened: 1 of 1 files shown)


LtdSauce:
Hi! I have not found tests, that actually check the cargo invokation and thus did not add any tests for this.
I have tested it locally though by executing it with this fix and saw that the toolchain file was not recognized before it, but after the patch it was.

This could be reproduced using any recipe that contains a toolchain, that uses a different channel/version than what is installed by default. I used the following for example:
[recipe.json](https://github.com/user-attachments/files/17333810/recipe.json)

PR #294, opened 2025-01-15
Upgrade libc crate

paolobarbolini:
While packaging this crate for Alpine Linux I've realized that locked dependencies haven't been upgraded for a while. In particular the current locked version of `libc` doesn't support loongarch64. This PR bumps it so that we can remove our fix downstream (https://gitlab.alpinelinux.org/alpine/aports/-/blob/a341ac413c530a2c4288486f2f16f1153d5c1f5e/testing/cargo-chef/APKBUILD#L26).


LukeMathWalker:
I did a general upgrade on main. Thanks for the reminder!

PR #303, opened 2025-04-03
Add a way to set the jobs limit

Bowarc:
Not a big thing but i needed it.

Maybe could be useful to others.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Context};
 use chef::{
-    AllFeatures, CommandArg, CookArgs, DefaultFeatures, OptimisationProfile, Recipe, TargetArgs,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -154,6 +155,9 @@ pub struct Cook {
     /// projects that rely on a custom build system (i.e. not `cargo`).
     #[clap(long)]
     no_build: bool,
+    /// Specify the number of worker that you want to use
```
reneleonhardt, line 158: Maybe more concise:
```suggestion
    /// Number of parallel jobs
```

reneleonhardt (commented):
I'm not a maintainer, just a suggestion 😄
Documentation would be nice for the user though...

Bowarc:
Ah yea, I forgot that I disabled it on save, mb

Bowarc:
Should be good now

PR #309, opened 2025-06-24
feat: Add --ignore-manifest cook arg

sharnoff:
One way to speed up builds involving large monomorphized dependencies is to separate their usage into a separate crate, which is then cached.

It's helpful if that can be a package in the same workspace, but AFAICT cargo-chef doesn't currently have a mechanism to extend its caching to some workspace members.

So to support that, this PR adds an `--ignore-manifest` flag to request that cargo-chef doesn't overwrite that `Cargo.toml` file or clean up the build artifacts from that package.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -115,6 +115,9 @@ pub struct Cook {
     /// Path to Cargo.toml
     #[arg(long)]
     manifest_path: Option<PathBuf>,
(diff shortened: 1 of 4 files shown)


LukeMathWalker:
Better support for large workspaces is definitely desirable, but this requires more upfront design.

PR #320, opened 2025-10-11
Add riscv64 architecture

eshattow:
Add riscv64 architecture to Docker and Publish Binaries workflows. Windows specific targets are not added.


eshattow:
Depends on #321

LukeMathWalker:
Thanks for the PR! Unfortunately, these targets are not mainstream enough within our userbase to be maintained upstream in my opinion. 
Users can build from source via `cargo install` should they need `cargo-chef` on those architectures.

Issue #323, opened 2025-11-16
Independent workspace members get rebuild unnecessarily


preiter93:
I've noticed that using cargo-chef independent workspace members get rebuilt unnecessarily.

## Problem

For example say we have a workspace
```sh
├── Cargo.toml
├── bar
└── foo
```

where `foo` and `bar` do not depend on each other, and foo depends on the "either" crate. If I change a dependency of foo (e.g. I upgrade its "either" version), `cargo-chef prepare --bin bar` still produces a recipe that references foo and its dependencies and thus the cache gets invalidated.

The recipe thats currently produced with
```
cargo-chef prepare --bin bar`
```
contains the manifests of foo and its dependencies in the lock_file:
```json
{
  "skeleton": {
    "manifests": [
      {
        "relative_path": "Cargo.toml",
        "contents": "[workspace]\nmembers = [\"bar\"]\nresolver = \"3\"\n[workspace.dependencies]\neither = \"1.15.0\""
      },
      {
        "relative_path": "bar/Cargo.toml",
        "contents": "[package]\nname = \"bar\"\nversion = \"0.0.1\"\n"
      },
      {
        "relative_path": "foo/Cargo.toml",
        "contents": "[package]\nname = \"foo\"\nversion = \"0.0.1\"\n\n[dependencies.either]\n"
      }
    ],
    "lock_file": "version = 4\n[[package]]\nname = \"bar\"\nversion = \"0.0.1\"\n[[package]]\nname = \"either\"\nversion = \"1.15.0\"[[package]]\nname = \"foo\"\nversion = \"0.0.1\"\ndependencies = [\"either\"]\n"
  }
}
```

As a result a change in foo’s dependencies invalidates the Docker cache for bar.

## Solution

I've drafted a recipe-reducer (https://github.com/preiter93/cargo-reduce-recipe) which reduces this down to
```json
{
  "skeleton": {
    "manifests": [
      {
        "relative_path": "Cargo.toml",
        "contents": "[workspace]\nmembers = [\"bar\"]\nresolver = \"3\"\n[workspace.dependencies]\neither = \"1.15.0\""
      },
      {
        "relative_path": "bar/Cargo.toml",
        "contents": "[package]\nname = \"bar\"\nversion = \"0.0.1\"\n"
      }
    ],
[...]

PR #324, opened 2025-11-23
feat(workspace): minimize manifests and lockfile for target members in recipe

preiter93:
For the `prepare` command, if a `--bin` member is specified:
  - Keep only the workspace manifests relevant to the target member
  - Filter the lock file to include only dependencies used by those manifests
  
  See https://github.com/LukeMathWalker/cargo-chef/issues/323

### DIFF
--- a/src/skeleton/mod.rs
+++ b/src/skeleton/mod.rs
@@ -11,7 +11,6 @@ use cargo_manifest::Product;
 use fs_err as fs;
 use globwalk::GlobWalkerBuilder;
 use guppy::graph::PackageGraph;
-use pathdiff::diff_paths;
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
> The chef cook is still missing workspace optimization. One thing is creating a minimal workspace Cargo.toml. And then one still has to create a optimized dockerfile to not copy unrelated members. I've done that [here](https://github.com/preiter93/rust-svelte-setup/tree/main/tools/docker-gen). I dont know if this could also become part of cargo chef. The dockerfile gen is a bit challenging for a general purpose library.

I'm not sure I see the issue here. If the recipe is minimized, then the re-hydrated workspace will be minimal and `cargo chef cook` will only "warm" the dependencies in the closure we computed. When you issue a `COPY . .` command, this will indeed copy everything, but the dependency cache will stay warm since it's a previous layer and the following `cargo build --bin <BIN_NAME>` command should only recompile the binary. Or am I missing something here?

LukeMathWalker:
> Looking through the code, I think we're making things more complicated then they need to be. In particular, the graph building part is not something I'd like to rebuilt into `cargo-chef`. Luckily enough, [`guppy`](https://docs.rs/guppy/latest/guppy/) can come to the rescue! Let's use it to resolve the dependency graph, and then walk that graph to find the package closure we need. It'll require reverting #211, but I think that's an acceptable trade-off.

I've migrated `cargo-chef` over to `guppy` in #336, so the groundwork should be there for you to reuse.

PR #364, opened 2026-06-10
ci: harden CI workflows with zizmor

### LINKED ISSUE #363: Harden CI workflows against script injection and credential exposure
## Problem

Pinning actions to SHAs (see [#361](https://github.com/LukeMathWalker/cargo-chef/issues/361)) removes one CI attack surface, but it does not address vulnerabilities inside the workflow definitions themselves.
The workflows in this repository contain a few well-known GitHub Actions anti-patterns that an attacker could abuse:

- **Template injection:** GitHub `${{ ... }}` expressions are expanded *before* the shell runs, so any expression interpolated directly into a `run:` block is substituted as raw text into the script. Values that an attacker can influence (matrix entries, step outputs, `cargo-dist` install commands, anything derived from a pull request) can therefore break out of the intended command and execute arbitrary shell with access to repository secrets and the `GITHUB_TOKEN`.
- **Credential persistence:** `actions/checkout` writes the `GITHUB_TOKEN` into `.git/config` by default (`persist-credentials: true`). Every later step - including any third-party action - can read that token and reuse it. This is called an [ArtiPACKED Vulnerability](https://thehackernews.com/2024/08/github-vulnerability-artipacked-exposes.html).
- **Excessive permissions:** the release workflow grants `contents: write` at the top level, so every job runs with write access even though only a couple of jobs actually need it.
- **Unpinned container images:** a job pulls a container image from a dynamic matrix, which cannot be expressed as a static SHA pin and needs an explicit, documented exception.

`cargo-chef` produces the DockerHub image and the released binaries that downstream build pipelines depend on. A compromise of these workflows would let an attacker tamper with those build artifacts. This supply-chain risk needs to be ruled out by security-aware users before adopting `cargo-chef`.

## Why static analysis (zizmor) fixes this

[`zizmor`](https://docs.zizmor.sh/) is a static analysis tool for GitHub Actions workflows. It detects template injection, credential persistence, over-broad permissions, unpinned images, and related findings, and many of them can be auto-corrected with `zizmor --fix`.

The fixes follow GitHub's own hardening guidance:

- The template-injection fixes use the recommended pattern of passing the expression through an [intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks) instead of interpolating it into the script body. This is the section `zizmor --fix` partly automated.
- The permission tightening follows ["Restricting permissions for tokens"](https://docs.github.com/en/actions/reference/security/secure-use#restricting-permissions-for-tokens), defaulting the workflow to `contents: read` and granting `contents: write` only to the jobs that publish.

## Proposed change

- Run `zizmor` as a dedicated CI job so new workflow changes are checked automatically.
- Apply `zizmor --fix` and the matching manual fixes:
  - move `${{ ... }}` expressions out of `run:` blocks and into `env:` variables that are referenced as quoted shell variables,
  - set `persist-credentials: false` on every `actions/checkout`,
  - default the release workflow to `contents: read` and escalate to `contents: write` per job,
  - document the one unavoidable `unpinned-images` finding with an inline `zizmor: ignore` and a justification.

## References

- GitHub Docs - [Good practices for mitigating script injection attacks](https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks)
- GitHub Docs - [Restricting permissions for tokens](https://docs.github.com/en/actions/reference/security/secure-use#restricting-permissions-for-tokens)
- GitHub Docs - [Using third-party actions](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)
- zizmor - [documentation](https://docs.zizmor.sh/)
- [`zizmorcore/zizmor-action`](https://github.com/zizmorcore/zizmor-action)

SvenPistre:
## PR content

This PR builds on (and includes the commits of because I cannot set my own branch as a target branch) #362. This PR hardens the workflow definitions themselves using [`zizmor`](https://docs.zizmor.sh/). It applies `zizmor --fix` plus the matching manual fixes, and adds a `zizmor` CI job so the workflows stay clean going forward.

The template-injection fixes follow GitHub's recommended pattern of passing each `${{ ... }}` expression through an [intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks) rather than interpolating it into the script body. The permission tightening follows ["Restricting permissions for tokens"](https://docs.github.com/en/actions/reference/security/secure-use#restricting-permissions-for-tokens).

This resolves #363.

## Findings addressed

| zizmor audit | Problem | Fix | Files |
| --- | --- | --- | --- |
| `template-injection` | `${{ ... }}` expressions interpolated into `run:` blocks are substituted as raw text and can execute attacker-influenced input as shell | Move each expression into an `env:` variable and reference it as a quoted shell variable; use `eval "$VAR"` for the dynamic `cargo-dist` install commands | `docker.yml`, `release.yml` |
| `artipacked` | `actions/checkout` persists the `GITHUB_TOKEN` to `.git/config`, where later steps can read it | `persist-credentials: false` on every checkout | `audit-on-push.yml`, `docker.yml`, `general.yml`, `scheduled-audit.yml` |
| `excessive-permissions` | the release workflow grants `contents: write` at the top level for all jobs | default to `contents: read`, grant `contents: write` only to the `plan` and `host` jobs that publish | `release.yml` |
| `unpinned-images` | a job pulls a container image from a dynamic matrix that cannot be SHA-pinned | documented inline `# zizmor: ignore[unpinned-images]` with a justification | `release.yml` |

## New CI job and action

A `zizmor` job is added to `general.yml` using [`zizmorcore/zizmor-action`](https://github.com/zizmorcore/zizmor-action), pinned to a full-length commit SHA like every other action.

| Action | Pinned version | Commit SHA | Verify (release / tag) |
| --- | --- | --- | --- |
| `zizmorcore/zizmor-action` | v0.5.6 | `5f14fd08f7cf1cb1609c1e344975f152c7ee938d` | <https://github.com/zizmorcore/zizmor-action/releases/tag/v0.5.6> |

## Notes

- No behavioural change is intended. The `eval "$VAR"` rewrites preserve the previous `cargo-dist` install behaviour while removing the direct expression interpolation.
- The `unpinned-images` exception is the only finding that is suppressed rather than fixed, because the image is provided by the `cargo-dist` generated matrix at runtime.

## References

- GitHub Docs - [Good practices for mitigating script injection attacks](https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks)
- GitHub Docs - [Restricting permissions for tokens](https://docs.github.com/en/actions/reference/security/secure-use#restricting-permissions-for-tokens)
- zizmor - [documentation](https://docs.zizmor.sh/) and [audit reference](https://docs.zizmor.sh/audits/)

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -14,16 +14,16 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
       - run: cargo test
 
   fmt:
     name: Rustfmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
           components: rustfmt
       - run: cargo fmt --all -- --check
@@ -32,8 +32,8 @@ jobs:
     name: Clippy
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
           components: clippy
       - run: cargo clippy -- -D warnings
@@ -43,12 +43,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
       - name: Install cargo-tarpaulin
-        uses: taiki-e/install-action@v2
+        uses: taiki-e/install-action@0631aa6515c7d545823c67cfae7ef4fc7f490154 # v2.81.8
         with:
           tool: cargo-tarpaulin
           checksum: true
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
       - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
       - run: cargo test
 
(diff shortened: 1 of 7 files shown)


.github/workflows/general.yml
```diff
@@ -43,14 +49,31 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+        with:
+          annotations: true
+          advanced-security: false
```
SvenPistre, line 79: If [Advanced Security](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security) is enabled for this repo, this can be removed. 
See https://github.com/zizmorcore/zizmor-action#advanced-security

